PR #10, opened 2020-02-08
Finally got rainbow dqn

Denys88:


mimoralea:
Just learning this paradigm of training on GPUs end-to-end, so forgive me if there is an obvious technical answer to this... but why no DQNs have been implemented? Maybe APEX architecture? Is this something you need help with?

Denys88:
@mimoralea I switched more to the robotics direction and we have SAC but I've found that it cannot utilize 16k envs, I could not make a config where I achieve better/faster results with more than ~1k envs compared  to the PPO. I have pretty good old DQN implementation on TF 1.X version  too.

PR #22, opened 2020-07-02
VDN 

schroederdewitt:
This PR includes Tarun's VDN implementation, alongside a partial implementation of PPO-S (where the bits left to do have to do with network architecture issues for the centralized critic)


common/experience.py
```diff
@@ -4,6 +4,67 @@
...
+class ReplayBufferCentralState(object):
+    def __init__(self, size):
+        """Create Replay buffer.
+        Parameters
+        ----------
+        size: int
+            Max number of transitions to store in the buffer. When the buffer
+            overflows the old memories are dropped.
+        """
+        self._storage = []
```
Denys88, line 16: I made measurements and found faster way to implement replay buffer with numpy only and preallocationg memory.
You can take a look into the common/experience.py

PR #40, opened 2020-11-22
Initialise RNN state during inference if one is needed and does not exist.

ArthurAllshire:


Denys88 (approved):

Denys88:
I thought I covered this case in Player. But if it is more convenient if you manually run inference I am fine.

ArthurAllshire:
Ah, I didn't notice the `init_rnn` function in player before. It is possible to do it by calling inference too, however it is more convenient from user side esp. during inference if you don't have to remember to make that call (and looks confusing if your model is not RNN).

PR #55, opened 2021-01-24
Tensorborad reward0 fix. Distributed training fixes.

ViktorM:


rl_games/common/a2c_common.py
```diff
@@ -148,7 +148,7 @@ def __init__(self, base_name, config):
             self.self_play_manager = SelfPlayManager(self.self_play_config, self.writer)
         
         # features
-        self.algo_observer = config['features']['observer']
+        self.algo_observer = config['features'].get('observer', None)
```
Denys88, line 151: Better to put EmptyObserver here or youll need to add checks in a lot of place in the code.

Denys88 (commented):
almost ok

PR #59, opened 2021-02-22
add logging passthrough of arbitrary scalar dict from env

ArthurAllshire:
This makes it easier to log eg. components of the reward which are hard to visualise otherwise.


Denys88:
Unfortunately it will not work that way.
IG is the unique case where we pass one one dict per N envs.
In case of the sc2 or atari envs we pass array of the dicts.
It means that dict[1] is expected to be results from env n2. 
@ViktorM  has code where he created his own custom observer.

PR #69, opened 2021-04-20
added value bootsrapping on time_out

NikitaRdn:
Added the bootstrapping of rewards with critic predictions on resets caused by time_outs. This greatly improves the training performance for envs where most resets are due to time outs. it is only applied if a 'time_out' tensor is returned in info, so should be compatible with all other envs.

### DIFF
--- a/rl_games/common/a2c_common.py
+++ b/rl_games/common/a2c_common.py
@@ -518,6 +518,9 @@ class A2CBase:
 
             self.obs, rewards, self.dones, infos = self.env_step(res_dict['action'])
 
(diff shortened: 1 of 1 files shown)


rl_games/common/a2c_common.py
```diff
@@ -518,6 +518,9 @@ def play_steps(self):
 
             self.obs, rewards, self.dones, infos = self.env_step(res_dict['action'])
 
+            if 'time_outs' in infos:
```
Denys88, line 521: could you add check if infos is Dict. It could have other type or None

PR #90, opened 2021-09-07
Statistics and naming improvements 

ViktorM:
- Added env (sim) step FPS reporting, renamed play and total FPS names, and few other tensorboard log variable names.
- Renamed lr_threshold -> kl_threshold
- Warning: older yaml training configs should be updated to using the new names,


ArthurAllshire:
Won't this break all previous configs...?

ViktorM:
@ArthurAllshire depending on where these configs are. For IGE it won't break as I've already updated them, as well as all the configs in rl_games repo itself. External people configs, as well as Isaac Gym Preview 2 one, will be broken for sure, but users can continue using an older rl_games version.

Denys88:
@ArthurAllshire  I've finally added pypi support. Current version is 1.0.2 https://pypi.org/project/rl-games/
This version will go as 1.1.0

PR #108, opened 2021-12-31
Fixes to get dict obs working

ArthurAllshire:


rl_games/common/a2c_common.py
```diff
@@ -620,7 +620,10 @@ def set_weights(self, weights):
     def _preproc_obs(self, obs_batch):
         if type(obs_batch) is dict:
             for k,v in obs_batch.items():
-                obs_batch[k] = self._preproc_obs(v)
+                if v.dtype == torch.uint8:
```
Denys88, line 623: should it go to the else block in the _preproc_obs call with tensor any way?

Denys88 (approved):

Denys88:
approved but not sure if we need commented change what I.

rl_games/common/a2c_common.py
ArthurAllshire: No, since it causes a crash if `normalise_input=True` (it expects `obs_batch` to be a dict if dict obs are used).

ArthurAllshire:
Replied in thread

ArthurAllshire:
^ Also applied the same fix in central value.

Denys88:
@ArthurAllshire happy new year :)
Can I merge it?

ArthurAllshire:
Ditto, yes I think you should be able to merge this 🎉

PR #110, opened 2022-01-03
Add option to not have vec env in environment

ArthurAllshire:
Needed when implementing some code that loads an actor but doesn't use it (for IL).

Feel free to reject this if there is a better way to load the network but I don't think that there is at the moment? Ideally we could have all the normalisation etc moved to the model class in which case there would be no need to load the whole agent in order to just use the actor network itself. Thoughts?


Denys88:
Ill think a little bit today after work.

Denys88 (approved):
looks good.

PR #119, opened 2022-01-31
Fix multi-gpu training with horovod.

ArthurAllshire:


Denys88:
@ArthurAllshire  if it is scaler for mixed precision I believe we can write wrapper and make sure it stores everything as pytorch tensors.

Issue #127, opened 2022-03-15
updates for brax_visualization.ipynb


boxiXia:
It seems brax changed their API and the [```brax_visualization.ipynb```](https://github.com/Denys88/rl_games/blob/master/brax_visualization.ipynb) needs some quick fix:
in cell#3, line#5:
```python
    config = runner.get_prebuilt_config()
```
needs to be commented/removed

in cell#3, line#8:
```python
env_config = config['env_config']
```
should change to
```
env_config = runner.params['config']['env_config']
```
In cell#5, line#14:
```env.state.qp``` should change to ```env.env._state.qp```

In cell#5, line#17:
```env.step(act.unsqueeze(0))``` should change to ```env.step(act)```

in cell#7, 
```python
display(visualize(env.env.sys, qps))
```
should change to
```python
display(visualize(env.env._env.sys, qps))
```

Denys88:
Thank you,
Ill test it. didn't touch it since their release :)

ViktorM:
Even after these changes, I still can't run it with the latest Brax:

ValueError: vmap got inconsistent sizes for array axes to be mapped:
the tree of axis sizes is:
(State(qp=QP(pos=1, rot=1, vel=1, ang=1), obs=1, reward=1, done=1, metrics={'reward_contact_cost': 1, 'reward_ctrl_cost': 1, 'reward_forward': 1, 'reward_survive': 1}, info={'first_obs': 1, 'first_qp': QP(pos=1, rot=1, vel=1, ang=1), 'steps': 1, 'truncation': 1}), 8)

@boxiXia what version of Brax did you use? Or maybe you have some other changes in place?

boxiXia:
I just did ```pip install brax``, brax version is: 0.0.10

ViktorM:
I did pip install from the local folder with Brax repo, after pulling the latest. It was also 0.0.10. But a cell with `qps.append(QP(env.env._state.qp))` still fails for me.

boxiXia:
Here is the "brax_visualization.ipynb" jupyter notebook modified for the visualization to work:
https://gist.github.com/boxiXia/1acd8897352a3404a172ee75d984fd20

boxiXia:
I used the master branch.@ViktorM

Denys88:
Looks like there is some difference between sac and ppo.
Ill make pr later today after work :) thanks @boxiXia !

ViktorM:
[...]

PR #129, opened 2022-03-17
Fixed brax_visualization

Denys88:
Fixed https://github.com/Denys88/rl_games/issues/127
@boxiXia and @ViktorM  should work for both ppo and sac now.

Also updated ppo_ant config with https://github.com/NVlabs/tiny-cuda-nn and got +10% speedup.


boxiXia:
Thanks for the update! I verified that both sac and visualization work. Thanks!

PR #130, opened 2022-03-27
Added flatten to be able to jit rl models

Denys88:
* Added jit example (trace model)
* Added onnx export/run example


rl_games/algos_torch/a2c_discrete.py
```diff
@@ -79,6 +79,19 @@ def update_epoch(self):
         return self.epoch_num
 
     def save(self, fn):
+
```
Denys88, line 82: temporary test, to be removed in the future

Issue #131, opened 2022-03-28
Multi-GPU with Central Value not working


ArthurAllshire:
Trying to run multi-gpu training with horovod, I get the following error:
```
[1,1]<stderr>:/opt/conda/lib/python3.8/site-packages/horovod/torch/sync_batch_norm.py:33: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
[1,1]<stderr>:  LooseVersion(torch.__version__) >= LooseVersion('1.5.0') and
[1,1]<stderr>:/opt/conda/lib/python3.8/site-packages/gym/spaces/box.py:84: UserWarning: WARN: Box bound precision lowered by casting to float32
[1,1]<stderr>:  logger.warn(f"Box bound precision lowered by casting to {self.dtype}")
[1,1]<stderr>:/workspace/isaacgymenvs/isaacgymenvs/tasks/allegro_hand.py:275: DeprecationWarning: an integer is required (got type isaacgym._bindings.linux-x86_64.gym_38.DofDriveMode).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
[1,1]<stderr>:  asset_options.default_dof_drive_mode = gymapi.DOF_MODE_POS
[1,1]<stderr>:/opt/conda/lib/python3.8/site-packages/horovod/common/util.py:227: DeprecationWarning: Parameter `average` has been replaced with `op` and will be removed in v0.21.0
[1,1]<stderr>:  warnings.warn('Parameter `average` has been replaced with `op` and will be removed in v0.21.0',
[1,1]<stderr>:Error executing job with overrides: ['task=AllegroHandLSTM', 'headless=True', 'multi_gpu=True', 'train.params.config.mixed_precision=False']
[1,1]<stderr>:Traceback (most recent call last):
[1,1]<stderr>:  File "train.py", line 137, in launch_rlg_hydra
[1,1]<stderr>:    runner.run({
[1,1]<stderr>:  File "/opt/conda/lib/python3.8/site-packages/rl_games/torch_runner.py", line 97, in run
[1,1]<stderr>:    self.run_train(args)
[1,1]<stderr>:  File "/opt/conda/lib/python3.8/site-packages/rl_games/torch_runner.py", line 78, in run_train
[1,1]<stderr>:    agent.train()
[1,1]<stderr>:  File "/opt/conda/lib/python3.8/site-packages/rl_games/common/a2c_common.py", line 1141, in train
[...]

PR #132, opened 2022-03-28
Fix multi gpu with CV

ArthurAllshire:


ArthurAllshire:
Fixes #131

Denys88 (approved):

ArthurAllshire:
Actually need to make some other changes to CV as it isnt actually being synced which breaks it, stand by :)

PR #135, opened 2022-04-02
Use vec env for inference

ArthurAllshire:
Not using it breaks inference in many cases if vec env API differs.

### DIFF
--- a/rl_games/common/player.py
+++ b/rl_games/common/player.py
@@ -4,6 +4,7 @@ import numpy as np
 import torch
 import copy
 from rl_games.common import env_configurations
+from rl_games.common import vecenv
 from rl_games.algos_torch import  model_builder
 
 class BasePlayer(object):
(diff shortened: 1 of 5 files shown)


rl_games/common/player.py
```diff
@@ -15,12 +15,12 @@ def __init__(self, params):
...
         if self.env_info is None:
-            self.env = self.create_env()
-            self.env_info = env_configurations.get_env_info(self.env)
+            self.env = vecenv.create_vec_env(self.env_name, self.config['num_actors'], **self.env_config)
+            self.env_info = self.env.get_env_info()
 
         self.value_size = self.env_info.get('value_size', 1)
         self.action_space = self.env_info['action_space']
-        self.num_agents = self.env_info['agents']
+        self.num_agents = self.env_info.get('agents', 1)
```
Denys88, line 23: could you check some openai gym envs and confirm it works with 1 or more envs.

rl_games/common/player.py
```diff
@@ -15,12 +15,12 @@ def __init__(self, params):
         self.clip_actions = config.get('clip_actions', True)
 
         if self.env_info is None:
-            self.env = self.create_env()
-            self.env_info = env_configurations.get_env_info(self.env)
+            self.env = vecenv.create_vec_env(self.env_name, self.config['num_actors'], **self.env_config)
```
Denys88, line 18: Need to think how to make it work with old configs  from openai gym.
Could you test if its ok to open 16 windows with atari for example?
Also it would be nice to be able to override the value from the player config (self.player_config).
In this case we can choose different number of actors if it was set.

ArthurAllshire: What do you mean by `old configs from OpenAI Gym`?
If we can't create environments with vec env the API is different between test and training time, and if you want to do wrapping of environments it makes it difficult.

ViktorM: Where are we at with this change? @ArthurAllshire did you have a chance to test it with non-IG envs and verify it doesn't break anything?

ViktorM:
Closed as vecenv for inference is already supported in 1.6.0

ViktorM:
Closed as vecenv for inference is already supported in 1.6.0

PR #138, opened 2022-04-08
Added proper seed

Denys88:
Most of the envs are still not reproducible.
Also added a few mujoco envs.


rl_games/configs/atari/ppo_breakout_torch.yaml
```diff
@@ -74,9 +74,9 @@ params:
...
-
+      seed: 5
     player:
-      render: True
+      render: False
       games_num: 200
       n_game_life: 5
-      determenistic: True
\ No newline at end of file
+      determenistic: False
```
ViktorM, line 82: Why false in a player?

Denys88: breakout has pretty weird use case where you can miss your shots 1000 times in a row, if set deterministic to false it will make other decision eventually. without big neural network with lstm from IMPALA paper it cannot be solved.

PR #144, opened 2022-04-19
add epoch/max epoch to print

erwincoumans:


Denys88 (approved):

Denys88:
@erwincoumans  merging it :)

PR #158, opened 2022-05-20
Prototype torch.distributed integration

vwxyzjn:
This PR Prototype `torch.distributed` integration for multi GPU


rl_games/common/a2c_common.py
```diff
@@ -966,7 +973,7 @@ def train(self):
                 update_time = 0
             if self.multi_gpu:
                     should_exit_t = torch.tensor(should_exit).float()
-                    self.hvd.broadcast_value(should_exit_t, 'should_exit')
+                    # self.hvd.broadcast_value(should_exit_t, 'should_exit') # what is the purpose of this?
```
Denys88, line 976: here is a chance that one job will exit  a little bit earlier, as result other jobs will crash.

rl_games/configs/atari/ppo_breakout_torch_multigpu.yaml
```diff
@@ -0,0 +1,84 @@
+params:
```
Denys88, line 1: btw could you use breakout from envpool?

rl_games/common/a2c_common.py
vwxyzjn: This method seems to be deprecated by horovod: the closest method I can find is https://horovod.readthedocs.io/en/stable/api.html#horovod.torch.broadcast_. 

What exactly is this `broadcast_value` doing? If rank 0's `should_exit_t=False` and rank 1's `should_exit_t=True`, would `broadcast_value` overwrite rank 1's `should_exit_t`?

rl_games/configs/atari/ppo_breakout_torch_multigpu.yaml
vwxyzjn: Envpool is actually more tricky. When prototyping with envpool it's actually slower with multi-GPU, at least out of the box. This is because envpool uses different threads and have complex interactions with these threads that are a bit difficult to control. For this reason, I have chosen the regular gym API for controlled performance.

PR #159, opened 2022-05-21
Added mini batch size per env

Denys88:
Added new parameter 'minibatch_size_per_env'. in this case minibatch_size = minibatch_size_per_env * num_actors.
If both 'minibatch_size' and 'minibatch_size_per_env' are here 'minibatch_size' will be used.

### DIFF
--- a/rl_games/algos_torch/central_value.py
+++ b/rl_games/algos_torch/central_value.py
@@ -45,8 +45,10 @@ class CentralValueTrain(nn.Module):
             self.scheduler = schedulers.IdentityScheduler()
         
         self.mini_epoch = config['mini_epochs']
-        self.mini_batch = config['minibatch_size']
-        self.num_minibatches = self.horizon_length * self.num_actors // self.mini_batch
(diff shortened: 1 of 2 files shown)


rl_games/algos_torch/central_value.py
```diff
@@ -45,8 +45,10 @@ def __init__(self, state_shape, value_size, ppo_device, num_agents, horizon_leng
             self.scheduler = schedulers.IdentityScheduler()
         
         self.mini_epoch = config['mini_epochs']
-        self.mini_batch = config['minibatch_size']
-        self.num_minibatches = self.horizon_length * self.num_actors // self.mini_batch
+        assert(('minibatch_size_per_env' in self.config) or ('minibatch_size' in self.config))
+        self.minibatch_size_per_env = self.config.get('minibatch_size_per_env', 0)
```
ViktorM, line 49: Shouldn't it be 1? In another case you could have division by 0 at line 51.

PR #162, opened 2022-05-23
Fix multi-gpu seeds

vwxyzjn:
This PR properly fixes the multi-GPU seeds.

It's a simple fix but quite ugly: a better fix probably would involved a larger refactoring such as putting horovod stuff into the main `runner.py`

### DIFF
--- a/rl_games/torch_runner.py
+++ b/rl_games/torch_runner.py
@@ -1,3 +1,4 @@
+import time
 import numpy as np
 import random
 import copy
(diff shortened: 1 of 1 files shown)


rl_games/torch_runner.py
```diff
@@ -55,20 +56,34 @@ def reset(self):
...
             torch.cuda.manual_seed_all(self.seed)
             np.random.seed(self.seed)
             random.seed(self.seed)
+            
+            # deal with environment specific seed if applicable
             if 'env_config' in params['config']:
                 if not 'seed' in params['config']['env_config']:
                     params['config']['env_config']['seed'] = self.seed
+                else:
+                    if params["config"].get('multi_gpu', False):
```
Denys88, line 85: did you mean true here?

Denys88: ah yes. I read it in the wrong way like if multi gpu is false then.

PR #165, opened 2022-05-25
Prototype torch.distributed

vwxyzjn:
Follow up on #158... Weird git issues.

This is a draft.


vwxyzjn:
Looking further into the bottlenect:

The left is CleanRL, right is rl-games. It seems the all_reduce cost is really high: maybe it's related to NN architecture? Will further look into it.

image

vwxyzjn:
Could be the architecture: I use the default one for `Ant`. After controlling the NN settings to be the same, they look more similar.

image

Issue #95, opened 2021-10-21
Multi-GPU usage


fil-mp:
How can I use multiple GPUs for simulation and training? I am enabling horovod but it seems that it can only use one device.

Denys88:
Could you show a whole error callstack?

Denys88:
I've found what causes this issue: only rank 0 process checked  number of epochs. Ill make a fix in a few days.

Denys88:
I've found 3 small issues:
1) exit on max epochs
2) exit on max rewards
3) uninitialized variable ( at least in discrete envs)
 I don't have access to the multigpu machine right now and simulated it with multi cpu. But it should work :)
https://pypi.org/project/rl-games/1.1.4/

1tac11:
Is this tested for multi instances or only multi GPUs on one instance? Since I had issues for multi instances.

ViktorM:
It was tested on a single node, up to 8 GPUs with Isaac Gym. Each instance of Isaac Gym was running on its own GPU.

PR #171, opened 2022-06-03
Deprecate `horovod` in favor of `torch.distributed`

vwxyzjn:
Follow up with #165 #158 

We did a benchmark with isaacgymenvs and `torch.distributed` shows consistently better scaling performance in `AllegroHand`.

Notable change - I disable the stats syncing, which does not seem to be that important to average across all workers at every step.

image

You can test it out with 

```
torchrun --standalone --nnodes=1 --nproc_per_node=2 runner.py --train --file rl_games/configs/ppo_cartpole.yaml
```

CC @ViktorM @markelsanz14


Denys88:
@vwxyzjn Ill take a look.
Strange but kl divergence syncing was needed to set right LR. Or you mean it is enough to calc it on the rank=0 gpu?

rl_games/common/a2c_common.py
```diff
@@ -958,10 +973,10 @@ def train(self):
                     print('MAX EPOCHS NUM!')
                     should_exit = True                              
                 update_time = 0
-            if self.multi_gpu:
-                    should_exit_t = torch.tensor(should_exit).float()
-                    self.hvd.broadcast_value(should_exit_t, 'should_exit')
-                    should_exit = should_exit_t.bool().item()
+            # if self.multi_gpu:
```
Denys88, line 976: I've found why I had this code: https://github.com/Denys88/rl_games/issues/95

PR #191, opened 2022-07-21
Fix multi-gpu bug with `central_value`

vwxyzjn:
This PR fixes multi-gpu bug w/ `central_value`.

Tested w/ `torchrun --standalone --nnodes=1 --nproc_per_node=2 train.py multi_gpu=True task=Ant task=ShadowHandOpenAI_LSTM num_envs=2048 wandb_activate=True wandb_entity=costa-huang train.params.config.minibatch_size=8192`

Which yields this [run](https://wandb.ai/costa-huang/isaacgymenvs/runs/373p80z0?workspace=user-costa-huang)

image

### DIFF
--- a/rl_games/algos_torch/central_value.py
+++ b/rl_games/algos_torch/central_value.py
@@ -1,3 +1,4 @@
+import os
 import torch
 from torch import nn
 import torch.distributed as dist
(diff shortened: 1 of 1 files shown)


rl_games/algos_torch/central_value.py
```diff
@@ -231,19 +243,24 @@ def calc_gradients(self, batch):
...
         loss.backward()
 
-        #TODO: Refactor this ugliest code of they year
+        # batch allreduce ops: see https://github.com/entity-neural-network/incubator/pull/220
+        all_grads_list = []
+        for param in self.model.parameters():
+            if param.grad is not None:
+                all_grads_list.append(param.grad.view(-1))
+        all_grads = torch.cat(all_grads_list)
+        dist.all_reduce(all_grads, op=dist.ReduceOp.SUM)
```
Denys88, line 252: why do we need  to work with gradients manually?
Are there any prallel optimizer like we had in horovod?

Denys88: I will merge it now but wanna look into the torch distributed module later.

Denys88 (approved):

rl_games/algos_torch/central_value.py
vwxyzjn: See https://github.com/vwxyzjn/cleanrl/pull/162#issue-1206232923. Option 2 (implemented here) is faster than the pytorch recommended gradient averaging -

```
def average_gradients(model):
    size = float(dist.get_world_size())
    for param in model.parameters():
        dist.all_reduce(param.grad.data, op=dist.ReduceOp.SUM)
        param.grad.data /= size
```

ankurhanda: The trouble with `cat` is that for a bigger network, we will have OOM because it's concatenating all the grads on a single GPU i.e. we can only roughly train a model of size 2GB on 8 GPU V100 with 16GBs each.

PR #192, opened 2022-07-27
Evaluation worker feature

alex-petrenko:
This adds a new feature that allows real-time evaluation and visualization of agents during the training session.
The evaluation worker is supposed to run in a separate process from the training session and thus enables evaluation on a small number of agents (i.e. 1 or 64) which still leaves enough resources to train on thousands of agents.

This will be triggered in isaacgymenvs using a new flag. Alternatively player can be started with `evaluation=True dir_to_monitor=/some/experiment/dir/containing/checkpoints` to monitor an existing training session.

Player with evaluation=True will continuously monitor the experiment dir for new checkpoints, load them, and visualize a new policy.

### DIFF
--- a/rl_games/common/player.py
+++ b/rl_games/common/player.py
@@ -1,10 +1,17 @@
+import os
+import shutil
+import threading
 import time
 import gym
 import numpy as np
 import torch
 import copy
+from os.path import basename
+from typing import Optional
+
 from rl_games.common import env_configurations
 from rl_games.algos_torch import  model_builder
+
 
 class BasePlayer(object):
     def __init__(self, params):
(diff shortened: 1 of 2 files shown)


alex-petrenko:
@ViktorM FYI

Denys88:
@alex-petrenko  will it work with IG on one gpu? as I remember I cannot create second IG on same gpu anyway?

alex-petrenko:
@Denys88 it worked fine on my 1080Ti provided there's enough memory, although I only tried on my machine.

alex-petrenko:
@ViktorM this is the version we'll need to use for the demo

Denys88:
@alex-petrenko  please let me know if you are going to add a few more changes or I can just merge it and refactor later.
Btw do you still need https://github.com/Denys88/rl_games/pull/195 this one?

alex-petrenko:
@Denys88 I think it's solid and works reliably. We were able to use it with both IGE and Omniverse IsaacGym.
It should be rather safe to merge since it does not do anything unless the evaluation flag is turned on.

If you don't want the file monitor thing (watchdog) to be in the main list of dependencies, you can remove it from setup py and add a warning that it should be installed under the evaluation section in the code.

PR #196, opened 2022-08-17
Added NVIDIA cule support

Denys88:
https://github.com/NVlabs/cule

Unfortunately it gets performance boost near 512-1024 envs and in general works slower thanenvpool because it is not required to have a lot of  envs to quickly solve atari.

### DIFF
--- a/rl_games/common/env_configurations.py
+++ b/rl_games/common/env_configurations.py
@@ -3,6 +3,7 @@ from rl_games.common import wrappers
 from rl_games.common import tr_helpers
 from rl_games.envs.brax import create_brax_env
 from rl_games.envs.envpool import create_envpool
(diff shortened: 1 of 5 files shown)


mimoralea:
Is this working 100%? I'd like to test it out if you can point me to the instructions.

Denys88:
I used https://github.com/NVlabs/cule/issues/34

mimoralea:
did you push a working Docker image by any chance, this library is proven difficult to build and run for me, not sure why.

Here is my attempt at running CuLE. https://github.com/mimoralea/cule/blob/master/envs/Dockerfile

This should build and run for the most part, but I still get error while running.

Here is the image: https://hub.docker.com/r/mimoralea/cule

`docker pull mimoralea/cule` almost there, but not quite!

Issue #237, opened 2023-04-24
Checkpoint when using SAC


Baerenstark2:
I'm using rl_games in through isaac gym and would prefer using SAC over a2c, but when creating a config file for SAC i noticed that it saves the new best states while learning but not the saves from the save_frequency parameter. This is a problem since, the saving of next best alway override itself, while with the saves after every x epochs all saves could be loaded and looked at after the finished training.
When looking through this repository I noticed, that the param eter save_frequencyis only read inside an a2c_common file, and now my question is, if is is simply not implemented for SAC(and if so if it could be added) or if it should work with SAC and I made some mistake?

ViktorM:
Hi @Baerenstark2,

Thanks for reporting an issue, I'll take a look.

PR #238, opened 2023-05-09
save_frequency works now with SAC as well

ViktorM:
https://github.com/Denys88/rl_games/issues/237

Fixed SAC not saving checkpoints with save_frequency, only the best ones.


Denys88 (approved):
add ticket to the description and lgtm

PR #240, opened 2023-06-03
Multi-node training support

ViktorM:
Random seeds and statistics are now handled correctly for multi-node training cases. No changes to training scripts are required. Multi-gpu/multi-node training is currently supported only for GPU-accelerated simulators like Isaac Gym.


Denys88 (approved):
Approved but please always fill description. What you did and how to run.

PR #246, opened 2023-07-19
add option for evaluating checkpoint

kellyguo11:
Adds an evaluation option to automatically search for checkpoints during training for inference use.

### DIFF
--- a/rl_games/common/player.py
+++ b/rl_games/common/player.py
@@ -1,8 +1,13 @@
+import os
+import shutil
+import threading
 import time
 import gym
 import numpy as np
 import torch
 import copy
+from os.path import basename
+from typing import Optional
 from rl_games.common import vecenv
 from rl_games.common import env_configurations
 from rl_games.algos_torch import model_builder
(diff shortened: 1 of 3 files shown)


Denys88:
Thanks for the PR, we will take a look on the weekends.

rl_games/common/player.py
```diff
@@ -71,6 +76,90 @@ def __init__(self, params):
...
+            self.file_events = PatternMatchingEventHandler(patterns)
+            self.file_events.on_created = self.on_file_created
+            self.file_events.on_modified = self.on_file_modified
+
+            self.file_observer = Observer()
+            self.file_observer.schedule(self.file_events, self.dir_to_monitor, recursive=False)
+            self.file_observer.start()
+
+    def wait_for_checkpoint(self):
+        if self.dir_to_monitor is None:
```
ViktorM, line 103: Maybe to check also if self.evaluation is true not if only the dir is not None?

`if self.evaluation and self.dir_to_monitor is not None:`

kellyguo11: fixed

ViktorM (approved):

PR #284, opened 2024-05-06
fixed bug with multi-gpu training for continuous a2c in a2c_common.py

tttonyalpha:
Fixed bug with missed torch.cuda.set_device(self.local_rank), which causes a problem when two different parallel processes try to use the same GPU in multi-gpu training with continuous a2c implementation

### DIFF
--- a/rl_games/common/a2c_common.py
+++ b/rl_games/common/a2c_common.py
@@ -1308,6 +1308,7 @@ class ContinuousA2CBase(A2CBase):
         self.curr_frames = self.batch_size_envs
 
         if self.multi_gpu:
+            torch.cuda.set_device(self.local_rank)
             print("====================broadcasting parameters")
             model_params = [self.model.state_dict()]
             dist.broadcast_object_list(model_params, 0)


ViktorM:
@tttonyalpha thank you for PR. Could you briefly describe what was an issues with running parallel processes on the same GPU?

ViktorM:
I just ran tests locally - training worked perfectly fine on a single GPU with running 2 different parallel processes.

tttonyalpha:
Sorry for a long reply: this problem only occurs when using multiple GPUs, when I tried to run continuous a2c training on 2 GPU using: `torchrun --standalone --nnodes=1 --nproc_per_node=2 ` I have got an error: `NCCL WARN Duplicate GPU detected : rank 1 and rank 0 both on CUDA device`, this occurs because according to (torch DDP documentation)[https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#distributeddataparallel] for N GPU parallelization you have to create N processes and each process must run exclusively on a single GPU, when we don't call `torch.cuda.set_device(self.local_rank)` before parameters broadcasting torch tries to use same GPU for each process (rank 0 and rank 1 in my case), which is incorrect

denysm88:
No problem, Thanks Ill merge it later this week.

ViktorM (approved):

PR #291, opened 2024-06-25
Fix SAC with input normalization

llinauer:
Atm, SAC does not work when giving normalize_input in the config.yml
I propose an easy fix:

1. In sac_agent.py, check if the self.normalize_input attribute is set. If yes, include the weights of the running_mean_std in the stat in get_weights method

2. In players.py, get self.normalize_input from the config and apply self.model.norm_obs to the observation in get_action
If normalize_input is False, then self.model.norm_obs will just be the Identity. If it is True, then it will apply the running mean to the observation

### DIFF
--- a/rl_games/algos_torch/players.py
+++ b/rl_games/algos_torch/players.py
@@ -199,7 +199,7 @@ class SACPlayer(BasePlayer):
         ]
 
         obs_shape = self.obs_shape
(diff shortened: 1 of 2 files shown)


Denys88 (approved):

Denys88:
Thanks! Looks good.

Issue #296, opened 2024-07-09
Potential Issues with Multi-GPU/Node Training with Central Network Weights Initialization


annan-tang:
Hi, Thank you for the great work on this project. I have a couple of questions regarding the multi-GPU/multi-node training implementation, specifically in the context of the central network.

From my understanding of the [source code](https://github.com/Denys88/rl_games/blob/7a2b25fc12ebf2a9ba9a05df56e210b35872ebb0/rl_games/common/a2c_common.py#L1322-L1324), it appears that the initial parameters of the actor_critic model on GPU rank_0 are broadcast to other GPU replicas to ensure they hold the same initial parameters. My questions are as follows:

Is broadcasting the initial parameters of the actor_critic model sufficient to ensure that all GPU replicas maintain the same parameters throughout training?
Given that different seeds might be used for each GPU, the central_network could potentially be initialized with different parameter weights. Could this be a potential issue for multi-GPU/multi-node training when using the central_network?
I appreciate your time and assistance in addressing these questions

PR #297, opened 2024-07-09
Ensure the consistency of central_value_net with the same initial parameters before training starts in a multi-GPU setting.

annan-tang:
Solution for Potential Issues with Multi-GPU/Node Training with Central Network Weights Initialization [#296](https://github.com/Denys88/rl_games/issues/296)

### DIFF
--- a/rl_games/common/a2c_common.py
+++ b/rl_games/common/a2c_common.py
@@ -1044,8 +1044,12 @@ class DiscreteA2CBase(A2CBase):
             torch.cuda.set_device(self.local_rank)
             print("====================broadcasting parameters")
             model_params = [self.model.state_dict()]
+            if self.has_central_value:
+                model_params.append(self.central_value_net.state_dict())
             dist.broadcast_object_list(model_params, 0)
             self.model.load_state_dict(model_params[0])
+            if self.has_central_value:
+                self.central_value_net.load_state_dict(model_params[1])
 
         while True:
             epoch_num = self.update_epoch()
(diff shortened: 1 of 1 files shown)


ViktorM:
Hi @annan-tang,

Thank you for PR, I'll take a look tomorrow. Could you please update it to the latest master?

annan-tang:
> Hi @annan-tang,
> 
> Thank you for PR, I'll take a look tomorrow. Could you please update it to the latest master?

Thank you very much, I will update it later. And I'm doing experiments to show the effectiveness. I will report more results later(within several days)

annan-tang:
Hi,

I conducted a comparison with and without the central value network initial parameters alignment code on a 2-GPU setting. I used the default Trifinger example in [IsaacGymEnvs](https://github.com/isaac-sim/IsaacGymEnvs/blob/main/isaacgymenvs/tasks/trifinger.py) with the following command:

```bash
torchrun --standalone --nnodes=1 --nproc_per_node=2 train.py multi_gpu=True task=Trifinger headless=True seed={xxx}
```

For each situation, I tested five groups of random seeds ({xxx}) and found that there is not much difference with and without the initial parameters alignment. The reward curves are illustrated below:

central_value_net_alignment

Based on these results, it appears that the initial parameters alignment has little effect on the 2-GPU setting. However, I'm not sure if this would change when scaling up to dozens of GPUs.

Denys88:
merging it.

PR #298, opened 2024-07-10
Update for tacsl release: CNN tower processing, critic weights loading and freezing.

iakinola23:
- enable loading the weights of the critic network from a PPO checkpoint, without the actor weights
- add flag to freeze critic while training actor
- adding ability to post-process the output of a conv tower with the spatial soft argmax or flatten layer

### DIFF
--- a/rl_games/common/a2c_common.py
+++ b/rl_games/common/a2c_common.py
@@ -643,6 +643,9 @@ class A2CBase(BaseAlgorithm):
             env_state = weights.get('env_state', None)
             self.vec_env.set_env_state(env_state)
 
+    def set_central_value_function_weights(self, weights):
+        self.central_value_net.load_state_dict(weights['assymetric_vf_nets'])
+
     def get_weights(self):
         state = self.get_stats_weights()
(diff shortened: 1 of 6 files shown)


rl_games/common/a2c_common.py
```diff
@@ -643,6 +643,9 @@ def set_full_state_weights(self, weights, set_epoch=True):
             env_state = weights.get('env_state', None)
             self.vec_env.set_env_state(env_state)
 
+    def set_central_value_function_weights(self, weights):
+        self.central_value_net.load_state_dict(weights['assymetric_vf_nets'])
```
ViktorM, line 647: A small mistype assymetric instead of asymmetric

iakinola23: This was to be consistent with the key defined [here](https://github.com/Denys88/rl_games/blob/master/rl_games/common/a2c_common.py#L636).

I'll go ahead and updated all three instances of `assymetric_vf_nets`.

ViktorM (approved):

PR #353, opened 2026-06-07
Add optional capability_manifest passthrough to checkpoints (#352)

idoco2003:
## What

Adds an optional `capability_manifest:` passthrough, as offered in #352.

When the training params carry a top-level `capability_manifest:` block, rl_games copies it **verbatim** into the saved checkpoint dict (both the A2C/PPO path and SAC). rl_games takes no position on its schema; the block is stored as-is and is a no-op when absent. Consumers read `checkpoint['capability_manifest']`.

```yaml
params:
  capability_manifest:        # optional, opaque to rl_games
    command_ranges:
      - { quantity: linear_velocity_x, min: -0.5, max: 1.0, unit: m_per_s }
    terrain_classes: [rigid, deformable]
  algo: { name: a2c_continuous }
  config: { ... }
```

This lets the limits a policy was trained under travel **with** the policy, instead of being transcribed by hand downstream and drifting.

## Changes
- `torch_ext.add_capability_manifest(state, params)` — small, schema-agnostic helper (verbatim copy, no-op when absent).
- `A2CBase` and `SACAgent` retain `self.params` and call the helper in `get_full_state_weights()`.
- `tests/test_capability_manifest.py` — covers present / absent / `None`-params.

Net diff is +64 lines, no behavior change for existing runs (the block is opt-in).

## Context
This comes from [URML](https://urml.dev) (an open, Apache-2.0 language for robot intent). The motivating discussion is #352. Prose is AI-assisted and maintainer-reviewed per URML's [VIBE.md](https://github.com/URML-MARS/URML/blob/main/VIBE.md); happy to adjust naming, placement (top-level params vs `config:`), or anything else to fit your preferences.

### DIFF
--- a/rl_games/algos_torch/sac_agent.py
+++ b/rl_games/algos_torch/sac_agent.py
@@ -19,6 +19,8 @@ class SACAgent(BaseAlgorithm):
     def __init__(self, base_name, params):
 
         self.config = config = params['config']
+        # Retain the full run params so an optional URML capability_manifest can travel with checkpoints.
+        self.params = params
         print(config)
 
         # TODO: Get obs shape and self.network
@@ -276,6 +278,9 @@ class SACAgent(BaseAlgorithm):
 
         if self.save_replay_buffer:
             state['replay_buffer'] = self.replay_buffer.state_dict()
+
+        # Optional URML capability manifest (no-op when absent), stored verbatim.
+        torch_ext.add_capability_manifest(state, self.params)
 
         return state
 
--- a/rl_games/algos_torch/torch_ext.py
+++ b/rl_games/algos_torch/torch_ext.py
@@ -87,6 +87,22 @@ def safe_load(filename):
         checkpoint = torch.load(filename, weights_only=False)
     return checkpoint
 
+def add_capability_manifest(state, params):
+    """Pass an optional URML capability manifest from the run params into a checkpoint, verbatim.
+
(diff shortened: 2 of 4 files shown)


Denys88 (approved):
Thanks @idoco2003 — this is clean and well-scoped. Verbatim, schema-agnostic passthrough, fully opt-in, no behavior change when absent, and no new imports (both agents already import `torch_ext`). `self.params` is a new attribute on `A2CBase`/`SACAgent` with no collision. Tests cover present/absent/None. Approving.

A couple of optional follow-ups (not blocking — happy to merge as-is):

1. **Survive resume.** `get_full_state_weights()` writes the manifest from `self.params`, but `set_full_state_weights()` never reads it back. So resuming from a manifest-carrying checkpoint with a config that omits `capability_manifest` would silently drop it from the next save. Usually fine (the manifest normally lives in the run params across resumes), but if you'd like it to round-trip regardless:
   ```python
   if self.params is not None and 'capability_manifest' in weights:
       self.params.setdefault('capability_manifest', weights['capability_manifest'])
   ```
   in both `set_full_state_weights()` paths — or just a sentence in the PR confirming the resume behavior is intentional.

2. **Scope note.** It's emitted from `get_full_state_weights()` (full checkpoint) only, not the weights-only `get_weights()` path — which I think is what you want; just confirming that matches the consumer story in #352.

Naming/placement (top-level `params:` vs `config:`) looks fine to me as-is.

PR #360, opened 2026-07-09
PPO/SAC hot-loop performance batch, sigma and RNN-state fixes, MJLab integration improvements

ViktorM:
This PR batches PPO/SAC hot-loop performance work with the correctness fixes it surfaced, plus MJLab (MuJoCo-Warp) integration with configs validated to full task success on GPU-simulation workloads.

**Performance defaults**

- **bf16 autocast on by default** on capable GPUs (`mixed_precision: false` to opt out). bf16 keeps fp32's exponent range, so **GradScaler is deleted entirely** — no loss scaling exists in any update path; updates are plain `backward()` → `clip_grad_norm_` → `step()`.
- **SAC `torch.compile` default `reduce-overhead`**: SAC's update loop is launch-overhead-bound and CUDA-graph replay is the measured win (+17–28% total fps depending on net size), while `default` mode loses to eager there. Only the submodules SAC actually calls are compiled, in place, so optimizer references and `state_dict` keys are unchanged. PPO stays on `default` (CUDA graphs are incompatible with RNN rollouts).

**Correctness fixes**

- **State-dependent sigma head init**: with `fixed_sigma: false`, `const_initializer` wrote the constant into the head's weight matrix, making the raw output `val · Σ(trunk activations)` — state-dependent garbage that NaNs exp heads and makes softplus heads over-explore from step one. The constant now initializes the **bias** with zeroed weights: uniform initial std, exactly the `fixed_sigma: true` semantics, with state-dependence learned. Latent for years because `fixed_sigma: true` is near-universal.
- **RNN state reset at episode boundaries**: `s[:, idx, :].zero_()` with a tensor index zeroes an advanced-indexing *copy* — a silent no-op, so LSTM hidden states leaked across episodes (visible as the multi-GPU ShadowHand LSTM regression). Now `s[:, idx, :] = 0` (`index_put_` scatter into the original), fixed for both the actor and central-value RNNs.
- **Episode-metrics logging chain**: empty-log key pinning, first-burst union-of-keys, and dict/tensor aliasing (log tensor values are cloned) — without these, per-term env metrics silently freeze or repeat.

**New config surface**

- Exploration shaping for state-dependent sigma: `min_sigma` (exploration floor), `sigma_parametrization: softplus`, `logstd_bounds`.
- `AdaptiveScheduler` gets explicit `min_lr` / `max_lr`. **Set these when using adaptive scheduling** — the legacy defaults (1e-6, 1e-2) let the LR walk to the ceiling on tasks whose measured KL sits below threshold.
- `vecenv_type`, `algo_observer`, `import_modules` config keys: external envs register from YAML alone (removed `run_mjlab.py`).

**MJLab integration + validated results** (same machine, reference batch geometry, full reference protocols)

- WujiHand in-hand reorientation (unmodified wuji-mjlab task): **17.1 vs 16.4** goal reaches/episode against the reference rsl-rl fork at equal training budget; the exported ONNX policy matches the officially released policy on the project's own sim2sim protocol (success 1.00).
- Go1 flat velocity: **86.8 vs 83.2** mean reward against the mjlab rsl-rl reference at its own 10k default budget and curriculum schedule.
- Lift-Cube-Yam: success **0.85 vs 0.72** against the reference recipe at the same frame budget.
- Shipped configs under `rl_games/configs/mjlab/` reproduce these numbers; comparison plots in `docs/MJLAB.md`.


rl_games/algos_torch/network_builder.py
```diff
@@ -318,6 +327,14 @@ def __init__(self, params, **kwargs):
                 mu_init(self.mu.weight)
                 if self.fixed_sigma:
                     sigma_init(self.sigma)
+                elif self.space_config['sigma_init'].get('name') == 'const_initializer':
+                    # state-dependent sigma head: the constant belongs in the
```
Denys88: This 6-line comment block is duplicated verbatim below (A2CResnetBuilder). Extract a helper, e.g. `init_state_dependent_sigma_head(sigma_layer, val)` — the name says everything the comment says, both blocks disappear, and the two builders can't drift apart.

rl_games/algos_torch/models.py
```diff
@@ -292,7 +292,21 @@ def forward(self, input_dict):
...
+            min_sigma = getattr(self.a2c_network, 'min_sigma', 0.0)
+            if getattr(self.a2c_network, 'sigma_parametrization', 'exp') == 'softplus':
+                sigma = torch.nn.functional.softplus(logstd) + min_sigma
+                logstd = torch.log(sigma)
+            else:
+                logstd_bounds = getattr(self.a2c_network, 'logstd_bounds', None)
+                if logstd_bounds is not None:
+                    logstd = torch.clamp(logstd, logstd_bounds[0], logstd_bounds[1])
+                sigma = torch.exp(logstd)
+                # optional exploration floor: sigma can never collapse below
```
Denys88: Same pattern: this whole if/else ladder wants to be a small function, e.g. `sigma, logstd = apply_sigma_parametrization(logstd, self.a2c_network)` — the exploration-floor comment then becomes a one-line docstring instead of prose inside `forward`.

rl_games/torch_runner.py
```diff
@@ -247,8 +266,12 @@ def run_train(self, args):
         else:
             # Parse configuration
             if isinstance(compile_config, bool):
-                actor_mode = "default"
-                critic_mode = "default"
+                # SAC's update loop is dominated by per-call launch overhead on
```
Denys88: The two SAC comment blocks here (+ line ~288) explain what a well-named helper would say for free: `actor_mode, critic_mode = default_compile_modes(agent)` and `compile_sac_submodules(agent, actor_mode, critic_mode)`. Keep the perf numbers in the PR description/changelog rather than in-line.

rl_games/envs/mjlab_vecenv.py
```diff
@@ -64,13 +66,27 @@ def step(self, actions):
 
         done = terminated | truncated
         info['time_outs'] = truncated
+        # surface the env's episode metrics (success rates, per-term rewards)
+        # so the observer logs them — reward totals alone hide task failure.
+        # mjlab reuses its extras dict across steps and emits an empty 'log'
```
Denys88: Shrink to one line or lift into a tiny `extract_episode_log(info)` helper. Also note `dict(info['log'])` is a shallow copy — if mjlab ever mutates the value tensors in place (rather than reassigning them), the aliasing bug returns; cloning tensor values would close it for good.

Denys88 (commented):
Ran the branch locally: 22/22 tests pass, and I reproduced the sigma-init bug on master vs the fix here (const val=-1, fixed_sigma false: master gives initial sigma in [8e-6, 1906] across states; this branch gives exactly exp(-1) uniformly) — good catch, the fix is right.

One question: `run_mjlab.py` used to call `wp.init()` explicitly before anything touched mjlab. With the launcher gone, does the `runner.py --file ...` path initialize Warp reliably (fresh process, not one that already had warp imported)? If mjlab handles it internally, all good — just confirming it wasn't masked by an already-initialized environment during validation.

Style ask across the diff (inline notes below): there are quite a few long comment blocks deep inside functions. Please either remove them or shrink them to one line — a small function with a good name always beats a comment deep inside the code, and half of these comments are describing exactly the function they want to become.

Denys88 (approved):
Approving — verified locally: 22/22 new tests pass, and I independently reproduced the sigma-init bug (master with const val=-1, fixed_sigma false: initial sigma spans [8e-6, 1906] across states; this branch: exactly exp(-1), uniform). The fix is correct and turns out to be load-bearing well beyond this PR: I ran a 7-variant policy-parametrization ablation on HalfCheetah/Ant/Walker2d/Hopper on top of this branch, and the best overall combination (existing continuous_a2c_tanh + state-dependent softplus sigma — wins 3/4 envs vs the fixed_sigma default, incl. Ant) is only usable *because of* the bias-init fix and the softplus/min_sigma options added here.

Non-blocking follow-ups from the earlier comments: the wp.init() question, comment blocks → small named functions, and a friendlier error for unknown algo_observer names. Also consider a future `init_sigma` key that solves the parametrization inverse internally (with softplus the init val must be softplus⁻¹(target), e.g. 0.5398 for sigma0=1 — easy to get wrong by hand).

Denys88:
Follow-up worth a shot on your 5090: I ran a policy-parametrization ablation on top of this branch (envpool gym-mujoco, 1000 epochs, sigma0 matched across variants, seed 42) and the existing-but-rarely-used `continuous_a2c_tanh` (the Brax NormalTanh port) + state-dependent sigma beats the fixed_sigma default on 3 of 4 envs:

| env | base (fixed sigma, clip) | tanh-squash + sd sigma | tanh-squash + fixed sigma | beta (Chou'17) |
|---|---|---|---|---|
| HalfCheetah | 1511 | 5385 | 5459 | **6150** |
| Ant | 4878 | **5184** | 4579 | 4560 |
| Walker2d | **5798** | 2987 | 4288 | 4222 |
| Hopper | 2018 | **3289** | 2551 | 1845 |

Naive tanh on mu (no Jacobian) is catastrophic (HC 1513) — the squashed log-probs are the whole story. All bias-free bounded policies gain +190–300% on HalfCheetah, i.e. wherever actions saturate. Notably the state-dependent variants only work at all because of this PR's sigma-head bias-init fix.

Since mjlab needs CUDA and I'm on Apple silicon, I prepared sigma0-matched configs you could drop in and run at the 5000-epoch protocol:
- `ppo_go1_velocity_tanh.yaml` / `ppo_g1_velocity_tanh.yaml`: `model: continuous_a2c_tanh`, `fixed_sigma: false`, `sigma_init val: 0.5398` (softplus-inverse of 0.999 -> sigma0 = 1.0, matching your current exp(0))
- `ppo_wujihand_reorient_tanh.yaml`: same model, `sigma_init val: -0.4307` (-> sigma0 ~ 0.50, matching your softplus(-1.05)+0.2 recipe), `bounds_loss_coef: 0`

One gap this exposed: `ModelA2CContinuousTanh` hardcodes `NormalTanhDistribution(...)` with `min_std=0.001`, so the WujiHand exploration floor of 0.2 can't be expressed with the squash — worth threading `min_sigma` from the space config into the distribution ctor (one line) so the tanh model composes with this PR's floor. Caveat on my numbers: single seed per cell, classic gym-mujoco — whether it transfers to mjlab's termination/curriculum structure is exactly what the Go1/G1 runs would tell us (Walker2d suggests termination-heavy locomotion may still favor the default; Ant suggests otherwise).

PR #361, opened 2026-07-12
Fix multi-discrete masked path (heterogeneous heads, CUDA) + DiscretizeActions wrapper

Denys88:
One real bug in the multi-discrete masked path (`ModelA2CMultiDiscrete.forward`, both branches): `np.split(action_masks, len(logits), axis=1)` splits into **equal** chunks — with heterogeneous head sizes (e.g. `[3,5,7]`) masks get mis-sliced and the forward crashes (`size of tensor a (5) must match b (3)`); with equal sizes it silently worked, which is why it survived. Fixed with `torch.split(masks, [l.shape[-1] for l in logits], dim=1)`.

Two claims from the original description were withdrawn after review (thanks @ViktorM): `np.split` on torch ≥ 1.8 tensors returns on-device torch views (numpy dispatches `swapaxes` to the tensor's own method), so there was no CUDA crash and no numpy round-trip; and batch-size-1 already worked bit-identically. The `squeeze(-1)` changes remain as explicitness/robustness only.

Also adds `common/wrappers.DiscretizeActions` (Box → Tuple of Discrete over a uniform grid, int or per-dim `bins`) — a discretized-control testbed the multi-discrete space never had. Rejects infinite Box bounds and emits actions in the Box dtype. Benchmarked against Gaussian PPO on envpool HalfCheetah (6 heads × 11 bins, matched hypers/seed): **1654 vs 1511** (adaptive LR) and **4588 vs 3715** (constant LR) — consistent with Tang & Agrawal 2020.

`tests/test_multidiscrete.py` covers: heterogeneous masked sampling + train path (both crash on master), batch-1 forward coverage, wrapper roundtrips (uniform + per-dim bins), infinite-bounds rejection, dtype passthrough.


ViktorM:
Approved after a doc correction. Verified on CPU + CUDA against master.

**Confirmed:**
- Heterogeneous heads (`[3,5,7]`) crash on master; the `torch.split` fix is
  numerically exact (matches renormalized log-softmax).
- `DiscretizeActions` trains end-to-end; clean merge; both suites green.

**Fix before merge — docs only, code is fine:**
Two of the three claimed defects don't repro on master:
- `np.split` on torch tensors returns on-device torch views (CUDA-crash claim
  only held for torch < 1.8); both call sites already pre-cast masks to torch.
- Batch-size-1 already worked (bit-identical outputs).

Reword `models.py:168,190`, the test docstrings, and release notes to claim
only the heterogeneous-heads fix.

**Nits:**
- `test_masks_stay_torch` duplicates the heterogeneous test under a wrong
  rationale.
- `_box_env` creates an unused, never-closed Pendulum env.
- `DiscretizeActions`: no finiteness guard on Box bounds (inf → nan actions);
  output hardcodes float32.

Denys88:
Review addressed in c2fb8d1 (thanks — the np.split-dispatches-to-tensor.swapaxes detail was a good catch, verified):
- comments/docstrings/PR body now claim only the heterogeneous-heads mis-slicing; CUDA-crash and batch-1 claims withdrawn, squeeze(-1) kept as explicitness only
- `test_masks_stay_torch` dropped; unused Pendulum env removed from the wrapper fixture
- `DiscretizeActions` now rejects infinite Box bounds and emits actions in the Box dtype (tests for both)

Also merged current master (post #362/#365) — full suite green locally (147 passed; the gloo test excluded only because sandboxed shells block its localhost rendezvous, CI runs it). Merging on green CI.

PR #362, opened 2026-07-12
PPO: mask next_step-autoreset garbage rows; linear sigma parametrization

ViktorM:
The PPO half of the autoreset correctness work (SAC half: b1ed755). On next_step-autoreset envs — envpool, native gymnasium 1.x vector envs — the reset step's row (action ignored by the env, filler reward, obs = the previous episode's terminal obs) entered PPO rollouts as real training data: one garbage loss sample and one garbage value target per episode boundary per env.

**Why loss-masking is the complete fix**: GAE's done-gated recursion already contains the contamination to that single row — the preceding `done` blocks leakage into the finished episode, and the new episode's first real row is computed correctly. Excluding the garbage row from the policy/value losses and advantage normalization is therefore mathematically sufficient; GAE itself is unchanged.

**Implementation**
- An explicit previous-step-dones tracker builds a validity mask during rollouts (the stored dones can't serve directly: their ones-initialization for the RNN fresh-state convention would flag the first row of training).
- The mask flows through the existing masks channel, which the losses, advantage normalization, KL reduction, and central-value dataset already honor. Episode reward/length accounting skips garbage rows.
- **Flagless**: activated by the env-declared `autoreset_mode`. same_step (Isaac-style) and ray paths produce no mask and are byte-identical to master. Multi-agent + next_step raises, mirroring SAC.

**Latent bugs fixed** (exposed by the first real producer on the masks channel):
- both algos consumed masks only when `is_rnn` — losses ignored a present mask on feed-forward nets (advantage normalization likewise)
- `bound_loss` returned int `0` without `bounds_loss_coef` (the masked path calls `.unsqueeze` on it), and the combined loss multiplied by a `None` coefficient

**Tests** (CPU, seconds): mask-vs-observation oracle on plain and RNN paths, gradient exclusion by poisoning (garbage rows poisoned to ±1e6 → bit-identical parameters after a full update), unpolluted episode stats, same_step no-mask guard, multi-agent guard.

**A/B evidence** (master vs branch, same-machine back-to-back pairs, mean of final window):

| Env | Seeds/side | master | branch | verdict |
|---|---|---|---|---|
| Pong (envpool) | 3 | 18.9 ± 1.3 | 19.2 ± 0.7 | parity (p=0.82) |
| Breakout (envpool) | 6 | 89.3 ± 13.9 | 90.9 ± 6.0 | parity (p=0.81) |
| CartPole (gymnasium native) | 10 | 332 ± 85 | 369 ± 62 | parity (p=0.31; early-curve marks also parity) |

No environment regresses; the excluded rows were estimator noise, and the branch reaches the same results training on strictly fewer (clean) samples. This lands as a correctness fix with no-harm evidence, not a performance claim.

---

**Also: `sigma_parametrization: scalar`**. The sigma head's raw output is the std itself, with a smooth softplus floor — under an entropy bonus the pressure on the raw output scales as 1/σ and self-limits, unlike log-space parametrizations where the same coefficient compounds exponentially. The floor must be smooth: a hard clamp's zero-gradient dead zone removes the restoring barrier (the first implementation NaN'd exactly this way; there's a regression test for the gradient below the floor). Validated end-to-end on Pendulum: reaches solved faster than the exp baseline (−176 vs −267 at epoch 100) and holds it (−149 vs −338 final-50). Opt-in; no default behavior changes.


Denys88:
Naming nit on the sigma parametrization: `scalar` reads as "a single shared sigma across action dims", which is not what this does — it's a per-dim identity/linear mapping with a smooth softplus floor. Suggest renaming to `linear` or `identity` before this lands, since config keys are hard to change later. (Also worth a doc note that `raw IS the std` only holds asymptotically: sigma = floor + softplus(raw − floor), so near the floor it deviates by up to ~ln 2.)

Denys88:
A few more substantive points from a full pass over the diff:

**1. Please split the multi-GPU stats sync into its own PR.** The PR description covers the masking and the sigma parametrization, but the branch also ships always-on cross-rank `RunningMeanStd` synchronization for every `multi_gpu` run, the `apply_masks` divisor change, the `schedule_type` rename, Brax removal, and the mjlab wrapper changes. The "same_step paths are byte-identical to master" claim holds for the masking feature but not for the PR as a whole — any multi-GPU same_step run (Isaac/mjlab, i.e. most users) now trains differently because of the stats sync. That change is also the one with the least evidence: the masking got poison tests + 3 A/B envs, while the stats sync has unit tests where all-reduce is simulated by `t.mul_(2)` and a docstring referencing a measured early-training deficit with no numbers. I'd like it as a separate PR with the actual A/B curves and an opt-out flag. One numerical note for that PR: reconstructing `sum_x2 = (var + mean²)·count` and recovering `var = sum_x2/n − mean²` is catastrophic cancellation when `mean² ≫ var` — fine while the buffers are float64, but that's an implicit dependency worth a comment.

**2. RNN + next_step: loss masking doesn't stop hidden-state contamination.** The garbage row's obs (previous episode's terminal obs) still flows through the LSTM forward pass inside its sequence, so the new episode's early hidden states are polluted before the first valid row — the mask can't fix that pathway. Probably minor, but the A/B evidence is feed-forward only (Pong/Breakout/CartPole) and the RNN test verifies mask placement, not learning parity. Either one RNN A/B run or an explicit known-limitation note in the docstring would close this.

**3. `_autoreset_prev_dones` is never cleared on re-reset.** It's initialized to `None` in `__init__` only. If `env_reset()` runs again (eval interleaving, custom loops), stale dones wrongly mask the first fresh row — and the comment ("None until the first step after env_reset") documents behavior the code doesn't implement. One-line fix: reset it in `env_reset`.

**4. Untested interaction: `value_bootstrap` + masking.** All masking tests run with `value_bootstrap: False`. The bootstrap term is added via reward shaping from `infos['time_outs']`; nothing proves a stale `time_outs` on the reset row can't inject a bogus bootstrap into the garbage reward that then sits in the GAE recursion. Cheap test to add.

Smaller:
- The `apply_masks` divisor "2.0.0 behavior change" framing can be softened: on master nothing produces `rnn_masks` (the channel is vestigial, always `None`), so this mostly activates dead code rather than changing existing RNN runs — worth confirming `central_value.py` has no producer either.
- Reported episode lengths on next_step envs shift by −1 (reset step no longer counted), so curves won't line up exactly against old runs — belongs in release notes.

The core masking work looks solid — the prev-dones tracker rationale is right and the poison test is a genuinely strong invariant. The asks above are about what's riding along with it.

Denys88 (approved):

PR #364, opened 2026-08-06
Fix dict-obs normalization: RunningMeanStdObs was un-scriptable

ViktorM:
models.py wraps RunningMeanStdObs in torch.jit.script for dict observation spaces, but its forward had an unannotated input (TorchScript infers Tensor, then `input.items()` fails to compile) and indexed the ModuleDict with a variable key (unsupported in TorchScript). Net effect: dict-obs + normalize_input crashed at model construction -- single- and multi-GPU alike -- since the container was first scripted (#315).

Fix: annotate forward as Dict[str, Tensor] and iterate the ModuleDict instead of indexing it. Tests: scripted construction (the exact models.py usage), scripted-vs-eager bit-identical stats and outputs, denorm roundtrip.

### DIFF
--- a/rl_games/algos_torch/running_mean_std.py
+++ b/rl_games/algos_torch/running_mean_std.py
@@ -124,6 +124,11 @@ class RunningMeanStdObs(nn.Module):
         })
 
     def forward(self, input: Dict[str, torch.Tensor], denorm: bool = False) -> Dict[str, torch.Tensor]:
+        # loud key-mismatch guard: extra input keys would otherwise be
+        # silently dropped, masking an env/config mismatch (missing keys
+        # already fail loudly on the dict access below)
+        assert len(input) == len(self.running_mean_std), \
+            'observation dict keys do not match the normalizer keys'
         res: Dict[str, torch.Tensor] = {}
         for k, m in self.running_mean_std.items():
             res[k] = m(input[k], denorm)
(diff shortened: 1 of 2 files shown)


Denys88 (approved):
Reviewed — correct and surgical fix. Verified both TorchScript obstacles are real (`models.py:50` scripts the container unconditionally; unannotated `input` infers as `Tensor` so `.items()` fails, and variable-key `ModuleDict` indexing is unsupported), and the child `RunningMeanStd.forward` is already properly annotated, so the positional call `m(input[k], denorm)` is valid script. The regression test hits the exact `models.py` usage, and the bit-identical scripted-vs-eager stats assertion is a nice touch.

**Approving, but please fix one thing before merge**: switching iteration from `input.items()` to `self.running_mean_std.items()` changes key-mismatch behavior — *extra* input keys are now **silently dropped from the output dict** (previously a loud `KeyError`). That can mask an env/config mismatch: an obs key that never gets normalized and quietly disappears downstream. A one-line TorchScript-compatible guard makes it loud again:

```python
def forward(self, input: Dict[str, torch.Tensor], denorm: bool = False) -> Dict[str, torch.Tensor]:
    assert len(input) == len(self.running_mean_std)  # catch key mismatch loudly
    res: Dict[str, torch.Tensor] = {}
    for k, m in self.running_mean_std.items():
        res[k] = m(input[k], denorm)
    return res
```

(The flip side of the change — *missing* keys now raise instead of silently normalizing a subset — is an improvement; keep that.)

Non-blocking nits: `eager(data)[k]` is recomputed per key in the parity test (harmless in eval mode), and the stats comparison reaches into `RecursiveScriptModule` internals via `getattr` — works, just implementation-coupled.

PR #365, opened 2026-08-11
Revive RMS advantage normalization: crash, ignored mask, stale diagnostics

ViktorM:
normalize_rms_advantage has been dead since #315: the first statistics update assigned the plain tensors returned by update_moving_stats back to nn.Parameter attributes -- TypeError on epoch 1, every time. Behind that crash sat two more bugs it was masking: forward() accepted a mask (the RNN path passes rnn_masks) and silently ignored it, so padding rows would have biased the statistics; and PpoDiagnostics still read moving_mean/moving_var, attribute names from the class #231 replaced.

Fixes:
- GeneralizedMovingStats statistics move from requires_grad=False Parameters to buffers (same state_dict keys -- old checkpoints load) and the update relies on update_moving_stats' in-place mutation instead of reassigning.
- The mask is honored: masked-out rows are excluded from the update (bit-identical to feeding only the valid rows); an all-masked batch is a no-op.
- Diagnostics derive mean/var from _get_stats().

Tests: first-update crash regression + convergence (both mean_std impls), eval-mode no-update, masked-update bit-identity, all-masked no-op, state_dict key stability, diagnostics, and a full CPU PPO epoch with the feature enabled (the original crash site) -- 6 of 7 fail on unmodified master, all pass here.

### DIFF
--- a/rl_games/algos_torch/moving_mean_std.py
+++ b/rl_games/algos_torch/moving_mean_std.py
@@ -39,6 +39,16 @@ class GeneralizedMovingStats(nn.Module):
         else:
             raise NotImplementedError(self.impl)
 
+    def get_mean_std(self):
+        """Public accessor for the normalization statistics.
+
+        Returns (offset, invscale) as used for normalization. Note the
+        invscale/std is the EFFECTIVE value: variance is clamped to at
+        least 1/max^2 and eps is added before the square root, so it is
+        not the raw EMA variance (raw: sqrs - mean^2 for 'mean_std').
+        """
+        return self._get_stats()
+
     def _get_stats(self):
         if self.impl == 'off':
             return 0.0, 1.0
@@ -90,6 +100,11 @@ class GeneralizedMovingStats(nn.Module):
         return mean, sqrs
 
     def _update_stats(self, x, mask=None):
+        # EMA semantics: every update moves the statistics by the same decay
+        # regardless of how many (valid) rows the batch carries -- a 3-row
+        # masked batch weighs as much as a 3000-row one. Statistics are also
+        # rank-local under multi-GPU (not covered by the cross-rank sync of
+        # obs/value normalizers; see #363).
         if mask is not None:
             # honor the validity mask: masked-out rows (RNN padding, autoreset
             # filler) must not move the statistics
--- a/rl_games/algos_torch/moving_mean_std.py
+++ b/rl_games/algos_torch/moving_mean_std.py
@@ -19,30 +19,23 @@ class GeneralizedMovingStats(nn.Module):
         self.eps = eps
         self.perclo = perclo
         self.perchi = perchi
+        # statistics are buffers: not learnable, still checkpointed under the
+        # same state_dict keys as the former requires_grad=False Parameters
         if self.impl == 'off':
             pass
-        elif self.impl == 'mean_std':
-            self.step = torch.nn.Parameter(torch.ones((1), dtype=torch.int32), requires_grad=False)
-            self.mean = torch.nn.Parameter(torch.zeros((insize), dtype=torch.float32), requires_grad=False)
-            self.sqrs = torch.nn.Parameter(torch.zeros((insize), dtype=torch.float32), requires_grad=False)
-        elif self.impl == 'mean_std_corr':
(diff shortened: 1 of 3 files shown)


Denys88 (commented):
Reviewed against master — the core claims all verify. The crash diagnosis is correct (`self.mean = <plain tensor>` on an `nn.Parameter` attribute → `TypeError` in `nn.Module.__setattr__`, so the feature really was dead on epoch 1), the Parameter→buffer migration is checkpoint-compatible in both directions and `advantage_mean_std` never enters an optimizer, and the mask fix with the bit-identity test is exactly the right invariant. Test suite is strong and hits the real crash site. Nothing here blocks the bug fixes.

**Main ask: please run at least one real env before this merges.** All evidence in the PR is synthetic (`FakeNextStepVecEnv`, one CPU epoch, 16-unit MLP) — it proves the code executes, not that the revived feature still trains well. Since `normalize_rms_advantage` has been dead since #315, nobody has actually trained with EMA advantage normalization on the post-#231 class; its behavior differs meaningfully from per-batch normalization (momentum-lagged scale, non-zero-mean batches early, and `adv_rms_momentum` defaults to a fairly aggressive 0.5). A short A/B like the one #362 shipped — envpool HalfCheetah or Ant, 2–3 seeds, `normalize_rms_advantage: True` vs default per-batch — would show the flag is safe to recommend rather than merely non-crashing. Non-blocking for the fixes themselves, but I'd want it before the flag appears in any docs/configs.

Minor comments, none blocking:

1. **Diagnostics "var" is not the raw moving variance** — `_get_stats()` returns std after `clamp_min(var, 1/max²) + eps`, so the logged value is the floored/shifted variance. Fine for its purpose; either log `sqrs − mean²` directly or note it. Related: `_get_stats()` returns Python floats for `impl='off'`, where the new `.detach()` in diagnostics would crash — unreachable today since the agent always builds the default impl, but the dependency is silent.
2. **`_get_stats()` is private** — diagnostics reaching into it works but couples them; a small public accessor/property would be cleaner.
3. **Multi-GPU follow-up**: `_stats_sync_modules()` syncs obs/value normalizers across ranks, but the revived advantage EMA stays rank-local, so per-rank advantage scaling can drift under `multi_gpu`. It can't reuse the moment-merge (EMA state, no counts) so out of scope here — but worth an issue so the "normalizers are synced" invariant doesn't silently exclude this one.
4. **Doc nit**: after masking, each minibatch still moves the EMA with equal weight regardless of how many valid rows survived (10 valid rows count as much as 1000). Inherent to per-batch EMA and fine — one docstring sentence would preempt the question.

Denys88 (approved):
Approving — the crash, mask, and diagnostics fixes are verified correct and well-tested; none of my earlier comments block. The real-env A/B (envpool HalfCheetah/Ant, a couple of seeds) from my review still stands as a follow-up before recommending the flag in docs/configs.

Denys88:
Ran the real-env A/B requested in my review — envpool robotics tasks on this branch (ccf2bbf), CPU, 1000 epochs, 3 seeds per side, identical configs except `normalize_rms_advantage: true` + `adv_rms_momentum: 0.5`. Baselines re-run on the same branch (post-#362 autoreset masking active on both sides). Reported: mean of final-10% reward window.

| Env | baseline (per seed) | rms_advantage (per seed) | verdict |
|---|---|---|---|
| Ant-v4 | 4510 ± 267 (4534/4232/4764) | 4381 ± 108 (4384/4488/4271) | parity (p≈0.44) |
| HalfCheetah-v4 | 1482 ± 237 (1443/1736/1267) | 1564 ± 130 (1713/1481/1498) | parity (p≈0.60) |

Early learning: on HalfCheetah the EMA path is ahead on all 3 seeds at the 25% mark (1291 vs 1034 mean) and at 50% (1510 vs 1285); on Ant the curves are indistinguishable. One seed (HalfCheetah/123) showed a wobblier final window (±296 within-run) — nothing alarming, but consistent with momentum-lagged scale under a fast-moving advantage distribution.

Sanity: branch baseline at seed 42 matches the pre-#362 reference numbers on both envs, so the revived flag neither crashes nor regresses, and trains at parity with default per-batch normalization. That closes my main review ask — LGTM to merge.

PR #367, opened 2026-08-18
Vm/fix/release blockers

ViktorM:
Release-blocker batch: four silent-wrong-behavior fixes + notebook chore.
One commit per fix; every fix ships a regression test, and the seed/seq_len tests 
are proven to fail on unpatched master.

**`seed: 0` now actually seeds** — `if self.seed:` treated 0 as unset and
silently skipped torch/numpy/python/env seeding while logging that seed 0
was active. Guard on `is not None` (`None` still means "derive from time").

**`seq_len` fallback** — the deprecation warning implied the legacy key
still worked; it was discarded and RNNs silently trained with
`seq_length=4`. The legacy value is now used as a fallback (`seq_length`
wins when both are set), and the one shipped config still using the old
key (`ppo_pacman_torch_rnn.yaml`, 16) is migrated.

**SAC rejects `multi_gpu: true`** — SAC reads `RANK` only to gate the
writer; with no DDP wrapper or gradient collective, torchrun trained N
independent seed-offset agents racing on the same checkpoint files.
Raises `NotImplementedError` until real support lands (distributed
training is PPO-only).

**Adaptive-LR resume state** — the optimizer LR was checkpointed but
`last_lr`/`entropy_coef` were not, so the first `scheduler.update` after a
resume restarted the adaptive walk from the config LR (up to ~30× too
high after a late-training resume → KL spike). Both are now saved and
restored; old checkpoints without the keys keep config-derived values,
and multi-GPU composes with the existing rank-0 broadcast in `update_lr`.

**Chore:** the Colab notebook installs rl_games from `master` now that the
B1 branch is merged (flips to the PyPI pin at release).

175 tests green on the rebased tip.

### DIFF
--- a/notebooks/mjlab_training_colab.ipynb
+++ b/notebooks/mjlab_training_colab.ipynb
@@ -83,7 +83,7 @@
    "outputs": [],
    "source": [
     "# rl_games from git until the 2.0 PyPI release (then: %pip install rl-games mjlab mediapy)\n",
-    "RL_GAMES_REF = 'VM/feature/b1-ppo'   # flip to 'master' after the branch merges, PyPI after release\n",
+    "RL_GAMES_REF = 'master'   # flip to the PyPI pin after the 2.0 release\n",
     "\n",
     "import subprocess, sys\n",
     "subprocess.check_call([sys.executable, '-m', 'pip', 'install', '-q',\n",
--- a/rl_games/algos_torch/sac_agent.py
+++ b/rl_games/algos_torch/sac_agent.py
@@ -142,6 +142,13 @@ class SACAgent(BaseAlgorithm):
         # ────────────────────────────────────
         # Summary writer
         # ────────────────────────────────────
+        if config.get('multi_gpu', False):
+            # SAC has no gradient/parameter synchronization: torchrun would
+            # silently train independent seed-offset agents racing on the
+            # same checkpoint paths. Refuse rather than mistrain.
+            raise NotImplementedError(
+                'multi_gpu is not supported for SAC (no DDP/gradient sync); '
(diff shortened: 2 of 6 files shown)


Denys88 (approved):
Reviewed each fix against the code at head (2eabedf):

- **seed 0**: `load_config` sets `self.seed = params.get('seed', None)` and falls back to `int(time.time())` when None, so `is not None` makes the guard effectively always-true — seeding (torch/cuda/numpy/random + env_config propagation) now runs for seed 0 exactly like any other seed. No other falsy-seed checks remain in the runner.
- **seq_len fallback**: verified the deprecation-warning line in `a2c_common.py` is the only remaining `seq_len` reader in the package, so the nested `.get` covers the sole consumer path, and `ppo_pacman_torch_rnn.yaml` was the only shipped config on the legacy key.
- **SAC multi_gpu guard**: raise sits before any rank/writer setup; `Runner.load_config` only reads `multi_gpu` for rank bookkeeping (no dist init), so the error surfaces cleanly at agent construction. Test helper applies `config_overrides` after its `multi_gpu: False` default, so the test exercises the real path.
- **Adaptive-LR resume**: restore composes with `scheduler.update(self.last_lr, self.entropy_coef, ...)` on the next epoch and with the rank-0 broadcast in `update_lr`; both values are plain floats (no device baked into the checkpoint); old checkpoints keep config-derived values via `.get` fallbacks, and new checkpoints' extra keys are ignored by older loaders. SAC's own `get_full_state_weights` is untouched.

CI green (lint + 3.11/3.12). Nit, non-blocking: the notebook edit dropped the trailing newline at EOF.

PR #354, opened 2026-06-11
dm_control soccer 2v2 self-play with opponent league (EnvPool)

Denys88:
## Summary

Multi-agent **self-play training** for EnvPool's dm_control locomotion soccer envs (`BoxheadSoccer2v2-v1`, from Denys88/envpool#1): one shared PPO policy controls the home team across 256 parallel matches while an **8-type opponent league** plays the away team — scripted anchors (chaser, keeper, random, zero) plus **frozen past checkpoints** of the training policy (lagged self-play), auto-refreshed from the run dir.

```bash
python runner.py --train --file rl_games/configs/dm_control/boxhead_soccer_2v2_selfplay.yaml
python -m rl_games.envs.dmc_soccer_tools video --camera 3 --out match.mp4
python -m rl_games.envs.dmc_soccer_tools tournament --run-dir runs/boxhead_soccer_2v2_selfplay/nn
```

## What's included

- **`rl_games/envs/dmc_soccer_selfplay.py`** — vecenv adapter: players-as-actors batching (egocentric, team-relative obs + within-team one-hot player id), shaped rewards with dense-annealing
- **`rl_games/envs/dmc_soccer_opponents.py`** — opponent league + a minimal `FrozenPolicy` (rebuilds the actor from a checkpoint for cheap CPU inference; handles `_orig_mod.` torch.compile prefixes)
- **`rl_games/envs/dmc_soccer_tools.py`** — match video rendering + checkpoint round-robin tournament (goal-diff/ep cross table)
- **config + `docs/DMC_SOCCER_SELFPLAY.md`** — the validated recipe, documented as *the seven failure modes each choice prevents* (each observed and reproduced during tuning):
  1. concede penalty → ball-avoidance ("cowardice")
  2. dense stream ≥ goal reward + terminate-on-goal → dribble-farming
  3. two-sided ball-progress → punished by opponent attacks (learned helplessness)
  4. per-player chase rewards → everyone crowds the ball (fix: team-level chase)
  5. single-opponent training → late collapse (fix: the league)
  6. shaped-reward proxy plateau (fix: dense anneal → broke a 0.34→0.46 goals/ep plateau)
  7. entropy bonus + `fixed_sigma` → unbounded σ growth melts the policy into noise (fix: `entropy_coef: 0`)
- Registry entries follow the existing pattern (`_create_envpool`-style lazy import in `vecenv.py`, `env_configurations.py` entry)

## Validation

[...]

PR #368, opened 2026-08-21
dmc soccer selfplay

ViktorM:
DeepMind Control Soccer 2v2 self-play on envpool's native soccer environments (experimental, internal-research feature; not promoted in the README).

**Native envpool 1.2.7.** The adapter now targets `DmcSoccerBoxhead-v1` (also `DmcSoccerAnt-v1`, `DmcSoccerHumanoid-v1`) from envpool 1.2.7, released 2026-09-15, instead of the earlier fork build. `team_size`, `terminate_on_goal` and `time_limit` are `env_config` keys. Native facts the adapter is built on: per-player rewards with per-match termination; match blocks of `2 * team_size` contiguous rows, home team first; batches come back in thread-completion order (`info["env_id"]` is a changing permutation), so every batch is re-sorted before use (about 4% of a step); actions are always read env-major; goals pay +1/-1 per player and the score stats are per-step flags, so the adapter keeps its own score.

**Self-play.** Home team is the learner, away team is driven by a league of frozen snapshots plus scripted opponents; league reload on file mtime, seeded RNG, resume of the anneal and RNG state, next-step autoreset declared. A sign bug in the scripted opponents' steering (they turned away from the ball) is fixed.

**Install matrix.** `envpool>=1.2.7` on Python 3.12+, the 1.2.5 fallback on 3.11 unchanged; uv.lock relocked; CI's envpool resolution check expects 1.2.7 on 3.12+. Docs: DMC_SOCCER_SELFPLAY.md (native route, row contract), ENVPOOL.md (soccer ids; a one-line note that 1.2.7 also lists native `Mjlab-*` ids), README version note.

**Validation.** 43 soccer tests (stub with rotating block order, plus real-envpool tests for shapes at team_size 1 and 2, the permutation, home/away routing, env-state round trip, a scripted goal). Full suite 259 passed / 10 skipped on Python 3.12 with envpool 1.2.7 and on 3.11 with 1.2.5; the non-soccer subset is the master baseline. CPU smoke training and `--play` verified. The training results in the doc are from the fork build and are flagged as not yet reproduced on 1.2.7.


Denys88 (commented):
Read the full diff and surrounding code. Only the pre-1.2.7 fork source is available here, so the native 1.2.7 row contract itself is pinned only by your real-envpool tests; the adapter logic on top of it checks out. This supersedes my #354, which I'll close when this lands.

**Fix before merge**

- `dmc_soccer_opponents.py:98-108` (`FrozenPolicy.act`): a snapshot whose obs width differs from the live env (checkpoint from a different `team_size`, or a fork-era checkpoint left in the same `runs/.../nn` dir) raises a broadcasting error inside `league.actions`, which is outside the `_load` try/except (`:149-166`) — training crashes mid-run instead of the intended warn-and-keep. A `mean.shape[0] == obs_dim` check at load time routes it through the warn path.
- `tests/test_dmc_soccer.py:735-738` and `docs/DMC_SOCCER_SELFPLAY.md:151-153` read `stats_home_score`/`stats_away_score` as per-player "my team scored / conceded" flags (`home == (r > 0)`). In the fork source (`soccer.h:1161-1162`) they are team-global (`scoring_team == kHome` for every player). If 1.2.7 kept that, the real-envpool test fails on any match where the away striker scores first — passing once under seed 13 is weak evidence. The adapter doesn't use these flags, so this is doc/test only, but please confirm against the 1.2.7 source.
- `dmc_soccer_selfplay.py:212, 220`: `opponent` is never validated; a typo (`leauge`) silently falls into the `controlled = team_size` + uniform-random-away branch (`:354-358`).

**Decide**

- `pyproject.toml:290` `envpool>=1.2.7` on 3.12+: nothing outside the soccer adapter needs 1.2.7 (MyoSuite gate stays `>=1.2.6`; the adapter already fails loudly via the key assert at `:237-241`), yet every 3.12+ user re-running `pip install -e .[envpool]` gets upgraded and pulls the new 65 MB `envpool-assets-mjlab` wheel for a feature the README doesn't promote. The runtime assert might be enough. `uv.lock` is otherwise clean (only envpool + that one new wheel).
- Note CI never installs envpool, so the `@real_envpool` tests are always skipped there.

**Nits**

- League refresh: master's `torch.save` is non-atomic, so a refresh landing mid-write hits the "failed to load … keeping previous" WARNING and, since a later successful load clears the key (`:165`), re-warns on every collision. Functionally correct, just noisy.
- `sort_batch` does ~45 `_scatter` allocations per step (one per obs key incl. unused `stats_*`) plus the ~40-key concat; flattening first and scattering one `(rows, obs_dim)` array + the 3 stats keys would be cheaper. Not blocking.
- `check_player_layout` runs only at reset (`:338`); a wrong block layout with the right length on a step goes undetected.
- `dmc_soccer_tools.TeamController.act` (`:80`) has no 3-wide action guard and the tools have no `--env-name`, so they're BoxHead-only in practice; the doc table listing Ant/Humanoid for the tools is misleading.
- `docs/DMC_SOCCER_SELFPLAY.md:95-96` calls `max_episode_steps` "the adapter's own cap"; it's forwarded to `envpool.make_gymnasium` (`:231-234`).
- `boxhead_soccer_2v2_selfplay.yaml:424-427` `player.use_vecenv: True` is redundant (default already True for a registry entry without `env_creator`).

Verified fine: `sort_rows`/`sort_batch` ordering incl. rewards/terms and env-major action build; `team_size != 2` slicing; own score reset via `*= 1 - done`; `next_step` masking matches `a2c_common.py:393-402, 1109-1113` and `envs/envpool.py`; resume round-trips anneal + both RNGs; frozen opponents normalize with their own saved stats on CPU; steering sign verified from the BoxHead MJCF (steer hinge axis (0,0,−1), gear +55 → `steer = -2*ang` is correct); skip guard and stub tests run without envpool; `test_sac_correctness.py` change is just the wrapper-list grep test.

Merge order: overlaps #379 on `env_configurations.py`.

PR #369, opened 2026-08-24
Replace manual multi-GPU gradient sync with DistributedDataParallel

Denys88:
## Summary

Multi-GPU gradient sync now goes through `torch.nn.parallel.DistributedDataParallel` by default. The legacy flat all-reduce survives one release as `multi_gpu_grad_sync: 'flat_allreduce'` (single shared implementation in `torch_ext.flat_allreduce_grads`, used by the agents and the central value net), and `multi_gpu_scheduler_kl: 'global' | 'local'` controls the per-minibatch scheduler KL collective. Review round fixes: the CV DDP wrapper stays out of `state_dict()` (`__dict__` assignment), a 0-coef value-head term keeps DDP bucket accounting static when `has_value_loss` is False, and a hard guard raises if a custom agent never routes its training forward through `train_model()`. Tests on the 2-proc gloo harness (grad parity, unused-head config, state_dict invariance) pass on torch 2.7 and 2.13; CONFIG_PARAMS.md documents both knobs with measured tradeoffs.

- **Explicit lifecycle** — `setup_multi_gpu()`, called once at the start of `train()`, broadcasts initial weights from rank 0 (previously duplicated in the continuous and discrete `train()` bodies) and wraps the training forward in DDP: the agent's model directly, the central value net via `CentralValueTrain.setup_train_model()`. Two symmetric accessors name the intent at every forward call site: `inference_model()` (always the raw model — rollout, values) and `train_model()` (the DDP wrapper when one exists). Wrapping happens after checkpoint restore and `torch.compile`, so ordering is unchanged; checkpoints and attribute access go through the raw `self.model`, keeping `state_dict` keys unprefixed. `gradient_as_bucket_view=True` avoids extra copies; `broadcast_buffers=False` keeps running-stat buffers under the existing `multi_gpu_sync_stats` policy.
- **`multi_gpu_scheduler_kl: 'local'`** — the per-minibatch LR scheduler uses rank 0's local KL estimate instead of the cross-rank mean, dropping one blocking rendezvous per minibatch. The reduce only feeds rank 0's scheduler (other ranks run an Identity scheduler and receive lr through `update_lr`'s broadcast), so ranks never diverge.

## Motivation

Profiling 2-GPU PPO (IsaacLab `Isaac-Humanoid-Direct-v0`, 2× RTX PRO 6000 Blackwell, torch 2.7/cu128) showed NCCL kernels are only **~1.2% of GPU-busy time** for a default-size policy — the scaling loss comes from synchronization *structure*, not bandwidth: the manual gradient path exposes its all-reduce after backward (no overlap, two full-model copies per minibatch), and the per-minibatch KL reduce adds a blocking rendezvous where rank jitter accumulates.

## Benchmarks

2 × 16,384 envs, eager (`torch_compile=false`), 100 epochs, fps averaged over epochs 21–100.

**Default policy net `[400, 200, 100]` (~80 KB grads), 3 runs/arm:**

| arm | fps total (mean) | vs manual |
|---|---|---|
| manual sync (old master path) | 1,469k (±0.3%) | — |
| DDP + `scheduler_kl: 'local'` | **1,492k (±0.5%)** | **+1.6%** |
| DDP alone | 1,405k | −4% |

**Heavy net `[4096, 2048, 1024]` (~11M params / ~44 MB grads):**

| arm | fps total | scaling vs 1 GPU |
|---|---|---|
| 1 GPU | 291.5k | — |
| manual (old path) | 561.9k | 96.4% |
| DDP | **567.8k** | **97.4%** |
| DDP + `scheduler_kl: 'local'` | 566.2k | 97.1% |

Takeaways:
- For small control policies the old manual flat all-reduce was near-optimal — DDP's bucket/hook overhead can cost a few % there; pairing with `multi_gpu_scheduler_kl: 'local'` recovers it (+1.6% net). Standardizing on DDP drops ~40 lines of duplicated hand-rolled sync, composes with the PyTorch ecosystem, and wins as models grow.
- With heavy gradients DDP wins outright (+1.0%), and the gap should widen with larger models, more ranks, or slower interconnects, where the exposed manual all-reduce grows linearly while DDP hides it behind backward.
- `multi_gpu_scheduler_kl: 'local'` mainly helps small models where fixed rendezvous costs dominate; heavy compute amortizes it.

## Environment note (dual Blackwell)

Found while testing: torch 2.7's bundled **NCCL 2.26.2 crashes** (`illegal memory access` on the first collective, reproducible with a 10-line all-reduce) on a dual RTX PRO 6000 Blackwell workstation without NVLink; NCCL 2.29–2.31 hang probing PCIe P2P. Working combo: `nvidia-nccl-cu12==2.29.7` + `NCCL_P2P_DISABLE=1`. Unrelated to this PR's code but worth knowing for anyone running multi-GPU on similar hardware.

## Testing

All paths validated on 2 GPUs (torchrun, NCCL):
- **Continuous**: smoke + 100-epoch runs (converging rewards, unprefixed `state_dict` keys, checkpoint save/restore untouched); all benchmarks above collected with this implementation. Also verified under the default `torch.compile` wrapping (DDP wraps the compiled module fine on torch 2.7).
- **Central value + RNN**: `Isaac-Repose-Cube-Shadow-OpenAI-LSTM-Direct-v0` (asymmetric, LSTM policy + central value net) — both DDP wrappers engage and train.
- **Discrete**: `configs/ppo_cartpole.yaml` with `multi_gpu: True` — trains to reward ~400 (solved).
- **SAC**: never had multi-GPU gradient sync in this repo — nothing to migrate; unchanged.
- Single-GPU behavior is untouched (`train_model()` returns `self.model` as-is).

### DIFF
--- a/docs/CONFIG_PARAMS.md
+++ b/docs/CONFIG_PARAMS.md
@@ -124,6 +124,38 @@ behavior.
 A/B at 2 ranks (envpool Pong, 3 back-to-back seed pairs, 400 epochs):
 parity — pooled 19.46 ± 0.38 vs broadcast 18.96 ± 0.28, paired p = 0.398.
 
+### `multi_gpu_grad_sync`
+
+**Type:** str | **Default:** `'ddp'` | **Options:** `'ddp'`, `'flat_allreduce'` | **Applies:** multi-GPU (torchrun) PPO runs
+
+- `'ddp'`: gradients are averaged by `DistributedDataParallel` during
+  backward (bucketed, overlapped with compute). The training forward runs
+  through a lazily created DDP wrapper (`train_model()`); rollout inference,
+  checkpoints and attribute access keep using the raw model, so state_dict
+  keys are unchanged. The default.
+- `'flat_allreduce'`: the pre-2.x manual path — one flat all-reduce of all
+  gradients after backward. Kept for one release as a compatibility escape
+  hatch for downstream agents that override `calc_gradients()` and forward
+  through `self.model` directly; scheduled for removal.
+
+2-GPU A/B (Isaac Humanoid, 16k envs/rank): bit-identical gradients between
+modes; throughput within ±2% on a small policy net, DDP ahead ~1% at ~11M
+params. A multi-GPU run that never routes its training forward through
+`train_model()` under `'ddp'` raises at the optimizer step instead of
+silently training rank-divergent.
+
+### `multi_gpu_scheduler_kl`
+
+**Type:** str | **Default:** `'global'` | **Options:** `'global'`, `'local'` | **Applies:** multi-GPU runs with `schedule_type: per_minibatch`
+
+- `'global'`: the adaptive-LR scheduler sees the cross-rank mean KL (one
+  all-reduce per minibatch). The default and historical behavior.
+- `'local'`: the scheduler uses rank 0's local KL estimate, dropping one of
+  the per-minibatch collectives. The KL sample size the scheduler sees drops
+  by `world_size`; lr is broadcast from rank 0 either way, so ranks never
+  diverge. Worth trying when per-rank minibatches are large and profiling
+  shows the update phase rendezvous-bound.
+
 ### `capability_manifest`
 
 **Type:** any | **Default:** unset | **Applies:** PPO and SAC full-state checkpoints
--- a/docs/CONFIG_PARAMS.md
+++ b/docs/CONFIG_PARAMS.md
@@ -139,8 +139,12 @@ parity — pooled 19.46 ± 0.38 vs broadcast 18.96 ± 0.28, paired p = 0.398.
   through `self.model` directly; scheduled for removal.
 
 2-GPU A/B (Isaac Humanoid, 16k envs/rank): bit-identical gradients between
-modes; throughput within ±2% on a small policy net, DDP ahead ~1% at ~11M
-params. A multi-GPU run that never routes its training forward through
+modes. Throughput depends on model size: on the default small policy net
+`'flat_allreduce'` is faster (~4% eager, ~2.5% compiled — `DDP(compiled)`
+forfeits Dynamo's DDPOptimizer overlap), while at ~11M params `'ddp'` is
+ahead ~1% (97.4% scaling) and its advantage grows with model size and rank
+count. Pairing `'ddp'` with `multi_gpu_scheduler_kl: 'local'` recovers the
+small-net gap (+1.6% net over flat). A multi-GPU run that never routes its training forward through
 `train_model()` under `'ddp'` raises at the optimizer step instead of
 silently training rank-divergent.
 
--- a/rl_games/algos_torch/a2c_continuous.py
+++ b/rl_games/algos_torch/a2c_continuous.py
@@ -59,7 +59,7 @@ class A2CAgent(a2c_common.ContinuousA2CBase):
                 'seq_length': self.seq_length,
                 'normalize_value': self.normalize_value,
                 'network': self.central_value_config['network'],
-                'config': self.central_value_config,
+                'config': {**self.central_value_config, 'multi_gpu_grad_sync': self.multi_gpu_grad_sync},
                 'writter': self.writer,
                 'max_epochs': self.max_epochs,
                 'multi_gpu': self.multi_gpu,
@@ -119,7 +119,9 @@ class A2CAgent(a2c_common.ContinuousA2CBase):
                 self.clip_value
             )
         else:
-            c_loss = torch.zeros(1, device=self.ppo_device)
+            # 0-coef term keeps the value head in the autograd graph so DDP's
+            # static bucket accounting sees every parameter (exact-zero grads)
+            c_loss = 0.0 * values.sum() + torch.zeros(1, device=self.ppo_device)
         if self.bound_loss_type == 'regularisation':
             b_loss = self.reg_loss(mu)
         elif self.bound_loss_type == 'bound':
--- a/rl_games/algos_torch/a2c_continuous.py
+++ b/rl_games/algos_torch/a2c_continuous.py
@@ -170,26 +170,7 @@ class A2CAgent(a2c_common.ContinuousA2CBase):
             if self.zero_rnn_on_done:
                 batch_dict['dones'] = input_dict['dones']
 
-        # DDP wraps the model lazily on the first training step (after any
-        # checkpoint restore / torch.compile wrapping in torch_runner) and is
-        # used for the training forward ONLY: rollout inference, checkpoints
-        # and attribute access (a2c_network, running_mean_std, ...) keep going
-        # through self.model, so state_dict keys stay unprefixed.
-        # broadcast_buffers=False: running-stat buffers deliberately follow the
-        # multi_gpu_sync_stats policy instead of DDP's per-forward broadcast.
-        if self.multi_gpu and self.multi_gpu_ddp and not self._ddp_active:
-            from torch.nn.parallel import DistributedDataParallel as DDP
-            dev = torch.device(self.ppo_device)
-            self._ddp_model = DDP(
-                self.model,
-                device_ids=[dev.index] if dev.type == 'cuda' else None,
-                broadcast_buffers=False,
-                gradient_as_bucket_view=True,
-            )
-            self._ddp_active = True
-            print('Using DistributedDataParallel for gradient sync')
-
-        train_model = self._ddp_model if self._ddp_active else self.model
+        train_model = self.train_model()
         with torch.amp.autocast('cuda', enabled=self.mixed_precision, dtype=torch.bfloat16):
             res_dict = train_model(batch_dict)
             action_log_probs = res_dict['prev_neglogp']
--- a/rl_games/algos_torch/a2c_continuous.py
+++ b/rl_games/algos_torch/a2c_continuous.py
@@ -170,8 +170,28 @@ class A2CAgent(a2c_common.ContinuousA2CBase):
             if self.zero_rnn_on_done:
                 batch_dict['dones'] = input_dict['dones']
 
+        # DDP wraps the model lazily on the first training step (after any
+        # checkpoint restore / torch.compile wrapping in torch_runner) and is
+        # used for the training forward ONLY: rollout inference, checkpoints
+        # and attribute access (a2c_network, running_mean_std, ...) keep going
+        # through self.model, so state_dict keys stay unprefixed.
+        # broadcast_buffers=False: running-stat buffers deliberately follow the
+        # multi_gpu_sync_stats policy instead of DDP's per-forward broadcast.
+        if self.multi_gpu and self.multi_gpu_ddp and not self._ddp_active:
+            from torch.nn.parallel import DistributedDataParallel as DDP
+            dev = torch.device(self.ppo_device)
+            self._ddp_model = DDP(
+                self.model,
+                device_ids=[dev.index] if dev.type == 'cuda' else None,
+                broadcast_buffers=False,
+                gradient_as_bucket_view=True,
+            )
+            self._ddp_active = True
+            print('Using DistributedDataParallel for gradient sync')
+
+        train_model = self._ddp_model if self._ddp_active else self.model
         with torch.amp.autocast('cuda', enabled=self.mixed_precision, dtype=torch.bfloat16):
-            res_dict = self.model(batch_dict)
+            res_dict = train_model(batch_dict)
             action_log_probs = res_dict['prev_neglogp']
             values = res_dict['values']
             entropy = res_dict['entropy']
--- a/rl_games/algos_torch/a2c_continuous.py
+++ b/rl_games/algos_torch/a2c_continuous.py
@@ -118,10 +118,14 @@ class A2CAgent(a2c_common.ContinuousA2CBase):
                 return_batch,
                 self.clip_value
             )
-        else:
+        elif self._ddp_model is not None:
             # 0-coef term keeps the value head in the autograd graph so DDP's
-            # static bucket accounting sees every parameter (exact-zero grads)
+            # static bucket accounting sees every parameter (exact-zero grads).
+            # Only under DDP: it turns the value head's None grads into zeros,
+            # which lets optimizer weight_decay act on an otherwise dead head.
             c_loss = 0.0 * values.sum() + torch.zeros(1, device=self.ppo_device)
+        else:
+            c_loss = torch.zeros(1, device=self.ppo_device)
         if self.bound_loss_type == 'regularisation':
             b_loss = self.reg_loss(mu)
         elif self.bound_loss_type == 'bound':
--- a/rl_games/algos_torch/a2c_discrete.py
+++ b/rl_games/algos_torch/a2c_discrete.py
@@ -155,7 +155,7 @@ class DiscreteA2CAgent(a2c_common.DiscreteA2CBase):
                 batch_dict['dones'] = input_dict['dones']
 
         with torch.amp.autocast('cuda', enabled=self.mixed_precision, dtype=torch.bfloat16):
-            res_dict = self.model(batch_dict)
+            res_dict = self.train_model()(batch_dict)
             action_log_probs = res_dict['prev_neglogp']
             values = res_dict['values']
             entropy = res_dict['entropy']
--- a/rl_games/algos_torch/a2c_discrete.py
+++ b/rl_games/algos_torch/a2c_discrete.py
@@ -61,7 +61,7 @@ class DiscreteA2CAgent(a2c_common.DiscreteA2CBase):
                 'seq_length': self.seq_length,
                 'normalize_value': self.normalize_value,
                 'network': self.central_value_config['network'],
-                'config': self.central_value_config,
+                'config': {**self.central_value_config, 'multi_gpu_grad_sync': self.multi_gpu_grad_sync},
                 'writter': self.writer,
                 'max_epochs': self.max_epochs,
                 'multi_gpu': self.multi_gpu,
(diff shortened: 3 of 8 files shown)


ViktorM:
Using pre-historic torch 2.7 dosen't make any sense, at least 2.11+ should be tested.

Should we also keep the old implementation for compatibility reasons for now, but make DDP the default variant?

ViktorM:
Round 1's structural asks were taken well: discrete agents and the central
value now share the same `wrap_model_ddp` path (both former inconsistencies
gone), and the lazy-wrap design remains clean. But the update also removed
the opt-in flag AND the manual all-reduce entirely — DDP is now the only
multi-GPU path — which converts two round-1 risks into merge blockers.

**Verified on this tip (2-proc gloo, same-seed cross-version runs):**
- New-tip DDP vs current-master manual: first-minibatch synced grads and
  2-epoch final params **bit-identical** (0.0 max diff), clip engaged in
  both, state_dict keys unprefixed. The mechanism is right.
- compile-then-wrap (`DDP(OptimizedModule)`, our default
  `torch_compile: true`) still trains bit-identically.

**Blocker 1 — crash by default, no escape hatch (reproduced on this tip):**
central value + `use_experimental_cv: false` leaves the actor's value head
gradient-less; master trains it fine, this branch raises DDP's "Expected to
have finished reduction..." on the first backward. With the manual path
deleted there is no fallback config. Fix candidates: pass
`find_unused_parameters=True` when `has_value_loss` is False (cheap,
per-iteration overhead only for that config class), or `static_graph=True`,
or keep the value head in the loss with a 0-coef term.

**Blocker 2 — the only multi-GPU path is unmeasured:** our roadmap gated
any DDP switch on ≥2-GPU measurement, and under default
`torch_compile: true` the wrap order (`DDP(compiled)`) prevents Dynamo's
DDPOptimizer from engaging — so comm/compute overlap, DDP's perf
argument, is largely forfeited. Correctness is proven; the *speed* claim
is not. Before this merges as the sole path: same-window A/B, manual
(master) vs this branch, compiled and uncompiled, ≥2 GPUs. If it's not
faster compiled, consider wrapping before compile in torch_runner or
documenting `torch_compile: false` for multi-GPU.

**Still open from round 1:**
- `multi_gpu_defer_kl` naming: the reduce is skipped, not deferred —
  suggest `multi_gpu_scheduler_kl: 'global' | 'local'`; also note it
  divides the scheduler's KL sample size by world_size (per_minibatch's
  16k+ minibatch guidance becomes per-rank) and removes 1 of 3
  per-minibatch collectives, not all.
- No tests: the 2-proc gloo harness from test_multigpu_stats_sync.py drops
  in — DDP-vs-single-rank gradient parity + the unused-params config are
  each ~30 lines. Happy to contribute both.
- No CONFIG_PARAMS.md entries (convention from #363).

**Suggested landing shape:** keep DDP-as-default (evidence above supports
it), but (a) fix Blocker 1, (b) attach the ≥2-GPU timing before deleting
the manual path in the same PR — or keep the manual path one release as
`multi_gpu_grad_sync: 'ddp' (default) | 'flat_allreduce'`, name-the-mapping
style, and delete it in 2.1 once the numbers are in.

ViktorM:
Round 3 — consolidating with the Codex review and the new benchmarks.

**The central-value checkpoint blocker is confirmed** (independent minimal
repro): `CentralValueTrain` is an `nn.Module`, so `self._ddp_model = DDP(...)`
registers the wrapper as a child — `state_dict()` grows
`_ddp_model.module.*` duplicates and a fresh agent's strict restore fails
with "Unexpected key(s)". The actor side is immune (the agent class isn't a
Module). Cheapest fix: keep the wrapper out of module registration —
`self.__dict__['_ddp_model'] = wrap_model_ddp(...)` (or `object.__setattr__`)
— plus a train→save→fresh-restore regression test. The agent-side lazy-wrap
claim ("checkpoints unprefixed") remains true and verified.

**Correction accepted on the unused-params failure mode:** the first
backward completes; DDP raises at the *next* training forward when it finds
the prior reduction unfinished. Same blocker, same fix space — and per the
Codex point, a blanket `find_unused_parameters=True` needs validation
against the `get_aux_loss()` pattern (mark-ready-twice hazard), so the
value-head case should be handled deliberately (0-coef loss term keeps the
graph static and costs nothing) and tested together with aux losses.

**Custom-algorithm silent desync — agreed, and it upgrades the fallback
from preference to requirement:** `trancate_gradients_and_step()` no longer
reduces, and any downstream subclass overriding `calc_gradients()` that
forwards through `self.model` (a documented extension pattern) now takes
rank-local optimizer steps with no warning. Minimum viable guard, one line:
in `trancate_gradients_and_step`, `if self.multi_gpu and self._ddp_model is
None: raise RuntimeError('multi-GPU gradient sync now runs through DDP —
route your training forward through self.train_model()')` — a custom agent
that never wrapped cannot reach the optimizer step silently. Same guard
idea applies to `CentralValueTrain.train_on_batch()`.

**Benchmarks — thanks for adding them; they answer round 2's ask and raise
one new question:** on the default net, eager, DDP *alone* is **−4%** vs
the manual path — the +1.6% comes from `defer_kl`, which defaults off. As
the PR stands, a default-config multi-GPU user gets the −4% arm. Since
`defer_kl` changes scheduler semantics (rank-0-local KL), making it default
needs the estimator evidence first (the offline KL-stream replay from round
2 is the cheap decisive test). Until then the small-net default regression
argues for keeping the flat all-reduce available:
`multi_gpu_grad_sync: ddp | flat_allreduce`, DDP default — which also
solves the custom-algorithm compatibility for one release. Heavy-net
results (+1pp scaling) and the compiled-mode measurement stand as the case
for DDP being the right long-term sole path.

Revised landing checklist (Codex's six + one):
1. Keep DDP as default implementation.
2. Fix asymmetric unused params deliberately (0-coef term; validate with aux losses).
3. `multi_gpu_grad_sync` fallback enum for one release, or the hard guard above at minimum.
4. Keep the CV DDP wrapper out of serialized state + fresh-restore test.
5. Two-rank tests (gradient parity, unused-params config, CV restore) — happy to contribute all three on the #363 gloo harness.
6. Document `defer_kl` as local-KL approximation (`multi_gpu_scheduler_kl: global|local` naming) in CONFIG_PARAMS.
7. Resolve the default-net −4%: defer_kl estimator evidence → flip its default, or accept and document the tradeoff.

Denys88:
Thanks for the thorough review rounds — especially for running the gloo parity verification and the two independent repros. All seven checklist items are addressed on the branch (commits `3c59d05` + docs follow-up):

**1. CV checkpoint blocker — fixed.** The wrapper is now assigned via `self.__dict__['_ddp_model']` so `nn.Module.__setattr__` never registers it; `state_dict()` keys are invariant and a fresh instance strict-restores. Regression test included (`test_ddp_wrapper_stays_out_of_module_state_dict`), and the asymmetric LSTM task now trains 2-GPU with the CV wrapper engaged and checkpoints saving cleanly.

**2. Unused value head — fixed deliberately, as suggested.** `calc_losses` (continuous and discrete) keeps the value head in the graph with a 0-coef term when `has_value_loss` is False — exact-zero grads, static graph, no `find_unused_parameters` (so no mark-ready-twice hazard with `get_aux_loss()`). Gloo test runs 3 consecutive steps and asserts the value-head grads are exactly zero.

**3. Silent-desync guard — added**, in both `trancate_gradients_and_step` and the CV train step: multi-GPU + `'ddp'` mode with an unwrapped model raises `RuntimeError` pointing at `train_model()` before the optimizer can step.

**4. Fallback enum — added**: `multi_gpu_grad_sync: 'ddp' (default) | 'flat_allreduce'`, with the legacy path extracted to `torch_ext.flat_allreduce_grads` (single shared implementation, used by agents and CV). Documented as a one-release compatibility escape hatch.

**5. Tests** — `tests/test_multigpu_ddp.py` on the 2-proc gloo harness: DDP-vs-flat gradient parity (allclose 1e-6, cross-rank identical), the 0-coef unused-head config, and the state_dict invariance/restore test. Passing on **torch 2.7.0 and torch 2.13.0** (the 2.11+ ask). Your extensions on the #363 harness are very welcome on top.

**6. Naming/docs** — `multi_gpu_defer_kl` → `multi_gpu_scheduler_kl: 'global' (default) | 'local'`; CONFIG_PARAMS.md entries added for both knobs, including the KL-sample-size-÷-world_size caveat and the collective count (1 of 3 per-minibatch collectives, as you noted).

**7. The measurement blocker — measured.** Same-window 2-GPU A/B (2× RTX PRO 6000, Isaac Humanoid, 16,384 envs/rank, 100 epochs):

| net | mode | fps total |
|---|---|---|
| default `[400,200,100]`, compiled (`torch_compile: true`) | flat_allreduce | **1,449k** |
| default, compiled | ddp | 1,413k (−2.5%) |
| default, eager (3 runs) | flat | 1,469k |
| default, eager | ddp | 1,405k (−4%) / **1,492k with `scheduler_kl: 'local'`** (+1.6%) |
| heavy `[4096,2048,1024]` (~11M params), eager | flat | 561.9k (96.4% scaling) |
| heavy, eager | ddp | **567.8k (97.4% scaling)** |

So your read is confirmed: `DDP(compiled)` forfeits the DDPOptimizer overlap and small nets favor flat, while heavy nets favor DDP and the gap should widen with size/ranks. The tradeoff is now documented in CONFIG_PARAMS rather than hidden; `multi_gpu_scheduler_kl` stays `'global'` by default pending the estimator evidence you proposed (the offline KL-stream replay is a good next step), and perf-sensitive small-net users have the flat fallback for this release.

On torch versions: the unit tests run on 2.13; the training-level validation is on 2.7 because that's what the Isaac Sim 5.1 pip stack pins on the benchmark box — happy to re-run the training A/B on newer torch once there's an env for it.

Denys88:
Follow-up on the lifecycle concern from review: the lazy first-call wrap is gone. `setup_multi_gpu()` now runs once at the start of `train()` — it absorbs the initial-weights broadcast that was duplicated in the continuous/discrete `train()` bodies, then wraps the agent model and (via `CentralValueTrain.setup_train_model()`) the central value net. Forward call sites use symmetric accessors: `inference_model()` for rollout/values, `train_model()` for the training pass. Revalidated on 2 GPUs: gloo tests pass (torch 2.7 + 2.13), `ddp` and `flat_allreduce` modes train, asymmetric LSTM + CV wraps on both ranks. Commit `721a4c4`.

ViktorM:
Benchmark follow-up with a plot twist: while running the round-2/3
measurement matrix we found **GPU P2P was broken on the benchmark box**
(AMD-Vi IO_PAGE_FAULT blocking peer DMA — bare NCCL all_reduce hung; fixed
with `iommu=pt` + reboot). The PR's original benchmarks almost certainly ran
host-staged. With native P2P, the picture improves for this PR:

**MJLab Go1, 2×16384, RTX PRO 6000 pair, step FPS (total in parens):**

| arm | torch 2.7 eager | 2.13 eager | 2.13 compiled |
|---|---|---|---|
| manual all-reduce (master) | 1221k (936k) | 1223k (938k) | 1220k (932k) |
| DDP (this PR) | **1258k (966k)** | 1240k (957k) | 1233k (958k) |
| DDP + defer_kl | 1238k (953k) | 1235k (953k) | 1226k (949k) |

- **DDP beats manual in every cell** (+1.0–3.1% step, +2.0–3.3% total) —
  the sole-path decision is now measurement-backed under working P2P.
- **defer_kl trails plain DDP in all six comparisons** — with P2P the KL
  all-reduce is nearly free. Recommend default off; its niche is
  expensive-collective regimes (multi-node / degraded interconnect — which
  is also the regime the original PR benchmarks accidentally measured).
- **compiled ≈ eager for DDP** at this net size — round 2's wrap-order
  concern (Blocker 2) resolves empirically; no torch_runner change needed.

IsaacLab DDP sanity (2×16384 vs same-stack 1 GPU): Humanoid 1.89×,
Shadow-Cube 1.90× — ~95% weak-scaling with native P2P.

Practical note for anyone benchmarking multi-GPU on Threadripper boards:
check `dmesg` for AMD-Vi IO_PAGE_FAULT before trusting NCCL numbers.

Remaining from round 3: the unused-params crash fix (Blocker 1), the CV
checkpoint key leak, the custom-algo guard/fallback, tests, and the
CONFIG_PARAMS entries. With those, this is mergeable — the perf case is
settled.

ViktorM:
Completing the measurement picture (all P2P-native, RTX PRO 6000 pair):

**Heavy net** (MJLab Go1 2×16384, mlp `[4096,2048,1024]`, ~44 MB grads),
total FPS: manual 479k · DDP 482k · DDP+defer 486k · DDP compiled **494k**.
Sync arms within ±1.5%; compile finally earns its keep here (+3.1%) via
the update phase.

**Your original frame, corrected** (IsaacLab Humanoid 2×16384, 2.7 eager):
manual 1047k · DDP 1039k/1056k (two runs) · DDP+defer 1047k — an exact
three-way tie within ±0.8% run variance. The −4% DDP deficit in the PR
table was the broken-P2P interconnect, not DDP.

**Compile default**: single-GPU A/B shows compile is a ~1% net loss on
default-size nets (495k→491k) and +2% on the heavy net — we're flipping
`torch_compile` default to off for 2.0 with the large-net opt-in
documented (branch VM/fix/compile-default-off).

Consolidated verdict for this PR: DDP as sole path is safe everywhere
measured (+2–3% total on GPU-pipeline sims, exact tie on CPU-bound
IsaacLab, never meaningfully worse); `defer_kl` shows no P2P-era benefit
on 2 GPUs — keep it opt-in, scoped in docs to expensive-collective
regimes (multi-node), or drop it if you'd rather not carry the flag.
Remaining before merge is unchanged from round 3: unused-params fix, CV
checkpoint key leak, custom-algo guard, tests, CONFIG_PARAMS entries.

ViktorM:
Re-reviewed the head commit (721a4c4). The round-3 central-value checkpoint blocker is fixed and verified — wrapper in `__dict__`, state_dict keys invariant, checkpoints load in both directions, `--load_critic_only` unaffected. flat_allreduce fallback, `multi_gpu_scheduler_kl`, and dropping defer_kl entirely all check out. Parity tests pass on 2-rank gloo.

One remaining gap, verified empirically on a 2-rank gloo group: the `train_model()` guard does not catch a custom agent that overrides `calc_gradients()` and forwards through `self.model` while inheriting `train()` — `setup_multi_gpu()` ran, `_ddp_model` is non-None, the guard passes, but DDP hooks never fire and ranks silently diverge. Suggest: set a flag in the DDP wrapper's forward and assert it in `trancate_gradients_and_step` (clear the flag after the step), so a bypassed forward raises the same actionable error. Plus a test for the raise path — current tests cover parity/buckets/state_dict only.

Nits: (1) the unconditional 0-coef value-head term changes behavior for `has_value_loss=False` + `weight_decay>0` (dead value head now decays — e.g. ppo_connect4_self_play.yaml); gate it on has_value_loss. (2) `a2c_discrete.get_masked_action_values` still uses raw `self.model` — inconsistent with the new accessor convention. (3) `CentralValueTrain.train_on_batch` has no callers and bypasses both sync modes — delete or route through `train_model()`.

Denys88:
Thanks for the round-4 pass and for verifying the checkpoint fix and parity tests.

**Bypass guard — agreed, real gap.** `trancate_gradients_and_step` only asserts `_ddp_model is not None`, so a custom agent that overrides `calc_gradients()` and forwards through raw `self.model` while inheriting `train()` passes the guard with DDP hooks never firing. Going with your suggestion: `wrap_model_ddp`'s forward sets a seen-flag, `trancate_gradients_and_step` raises the same actionable error when multi-GPU DDP mode is on and the flag is unset, and the flag clears after the step. Plus a test for the raise path next to the parity tests.

**Nit 1 — agreed, with one precision.** The 0-coef term already lives in the `has_value_loss=False` branch, so the gate it needs is on DDP actually being active, not on `has_value_loss`: when `_ddp_model` is unset (single GPU, or `flat_allreduce`) that branch goes back to plain `torch.zeros(1)`, the dead value head keeps None grads, and `weight_decay` never touches it (your connect4 self-play case). The term is added only under DDP, where bucket accounting requires it and the decay side effect is unavoidable — will document that in the comment there.

**Nit 2 — agreed.** `get_masked_action_values` runs under `torch.no_grad()` at rollout time so raw `self.model` is behaviorally fine, but it should follow the accessor convention — switching to `inference_model()`.

**Nit 3 — agreed.** `train_on_batch` has no callers and bypasses both sync modes; deleting it.

Will push all four to this branch.

ViktorM (approved):

Denys88:
Pushed ba53eef with all four items from the round-4 pass:

- **Bypass guard**: `wrap_model_ddp` now returns a tracked DDP subclass whose forward sets `forward_seen`; `trancate_gradients_and_step` raises an actionable error when DDP sync is on but the step's training forward bypassed the wrapper, and clears the flag after each step. `test_bypassed_training_forward_raises_and_flag_clears` covers the raise path and the clear (gloo).
- **0-coef term scoped to DDP**: single-GPU and `flat_allreduce` restore None grads for the dead value head, so `weight_decay` no longer decays it; the term (and its decay side effect) exists only while `_ddp_model` is active, documented at the site.
- `get_masked_action_values` → `inference_model()` (rollout-time `no_grad`, behavior unchanged).
- `CentralValueTrain.train_on_batch` deleted (no callers).

Both multigpu suites green: 14 passed (`test_multigpu_ddp.py` + `test_multigpu_stats_sync.py`). CONFIG_PARAMS.md already documents `multi_gpu_grad_sync` / `multi_gpu_scheduler_kl` from the earlier round.

ViktorM:
CI red is a stale test, not the change: `test_perf_fixes.py::test_central_value_autocast_is_bf16` counts `dtype=torch.bfloat16` == 2 in central_value.py, and deleting the uncalled `train_on_batch` legitimately removed one of the two autocast sites. Update the assert to == 1 (or better: count only reachable code). Everything else in ba53eef checks out on inspection — the tracked-DDP `forward_seen` guard closes the bypass we verified empirically, the 0-coef term is correctly scoped to `_ddp_model is not None` with None-grad restoration elsewhere, and `train_on_batch` had no callers. With the test count fixed: ready.

PR #370, opened 2026-08-25
MyoSuite: envpool-vectorized configs, gymnasium wrapper compat, player max_steps (supersedes #336)

ViktorM:
Supersedes #336, rebased onto current master — @Denys88's commits preserved with authorship.

- Six MyoSuite PPO configs, now **envpool-vectorized** (`MyoSuite/<task>`, 64 actors; envpool ≥1.2.5 ships all 398 official task ids) — the single-env `myo_gym` path and wrapper fix stay for custom tasks.
- `OldGymWrapper` (the `myo_gym` path's space-flattening wrapper): MyoSuite is gymnasium-API via its `myosuite.utils` shim — the wrapper now passes gymnasium tuples through instead of squashing them to the legacy 4-tuple (legacy inner envs still tolerated); docstring corrected; stub-env tests.
- Configurable player `max_steps`; explicit adaptive-LR bounds; one mislabeled config renamed (finger pose).

Verified: full suite green; elbow/walk/reach train end-to-end through envpool on CPU. Hyperparameter tuning on real hardware follows post-merge.


Denys88 (approved):

ViktorM:
Findings from the first full training + native-eval pass on these configs (RTX PRO 6000 workstation, 64 envs):

**envpool MyoSuite has no reset randomization** — parallel envs and successive episodes are bit-identical (0.0 obs spread; native myosuite differs by 0.70 between resets, and there is no spec knob to enable it). The `*Random` variants therefore train fixed-goal: checkpoints reproduce training reward exactly inside envpool (elbow 755.3 vs 755.7) but partially or fully fail on the native randomized env (elbow 2/4 episodes; hand-pose LSTM 4/4). Filed upstream: sail-sg/envpool#432. Follow-up commit documents this in `docs/MYOSUITE.md` and the recipe going forward is: train in envpool, always evaluate finals on native myosuite.

**Config fixes incoming on this branch:**
- `ppo_myo_finger_pose.yaml` had `name: MyoHandReorient8` (copy-paste) — every run dir it produced was mislabeled.
- `ppo_myo_hand_reorient.yaml` actually trains `myoChallengeDieReorientP1-v0` — renamed to `ppo_myo_die_reorient.yaml`.
- 128 actors everywhere, minibatch 4096 (finger had a stray 128), budgets set from measured solve points: elbow 200, finger 300, reach 500, walk 1000, hand-pose LSTM 4000, die reorient 10000 (unsolved at 5000, still climbing).

Also of note: the `myoLegWalk` policy is exploration-noise-dependent — 13K stochastic training reward, but deterministic eval falls after ~170 steps even inside envpool. Worth keeping stochastic eval in the loop for locomotion tasks.

PR #371, opened 2026-08-30
Warm-start the obs normalizer count (normalize_input_init_count)

Denys88:
## Problem

The input `RunningMeanStd` starts with `count = 1`. The first training minibatch (thousands of rows) outweighs the zero-mean/unit-var prior thousands to one, so the running stats jump straight to the batch stats within the first few updates. The policy recomputed under these shifted stats is then far from the rollout policy that collected the data — through no fault of the optimizer, and independent of the learning rate.

Measured on envpool MyoSuite `myoHandPoseRandom-v0`: first-epoch KL of **~68** vs a `kl_threshold` of 0.008 (an LSTM policy shows the same mechanism at smaller scale, ~0.077). With `lr_schedule: adaptive` the spike slams the LR to `min_lr` within the first epochs; because steady-state KL then sits inside the adaptive band, the LR never recovers and the whole run trains at the floor. Disabling `normalize_input` drops the first-epoch KL to 0.78, isolating the normalizer as the cause.

## Fix

Seed the running-stat count so the unit prior has weight and early updates *nudge* the stats instead of overwriting them:

```yaml
config:
  normalize_input: true
  normalize_input_init_count: 81920   # optional; None/absent = auto
```

- **Default (`None`/absent): one PPO epoch of samples** — `mini_epochs * horizon_length * num_actors * num_agents`. Self-scales to the setup; the prior fades after roughly one epoch.
- **`1` restores the legacy cold start** (this PR changes the default behavior — flagged deliberately, see below).

## Measurements

First-epoch KL, transformer policy on `myoHandPoseRandom` (identical config, LR pinned at 5e-5):

| init_count | epoch-1 KL |
|---|---|
| 1 (legacy) | 68.7 |
| 10 × minibatch (20,480) | 8.76 |
| 1 epoch = default (81,920) | **3.13** |
| 100 × minibatch (204,800) | 1.87 |

KL decays to steady state within ~4 epochs in all cases; probe-horizon rewards unaffected.

## Notes

- Threaded through `a2c_continuous` and `a2c_discrete`; `RunningMeanStdObs` (dict obs) forwards the count to each per-key normalizer.
- Value normalization (`normalize_value`) has the same cold-start mechanically; left untouched here to keep the change scoped — can follow up if wanted.
- Unit tests in `tests/test_obs_norm_init_count.py`; e2e-smoked on envpool MyoSuite with the auto default.

https://claude.ai/code/session_0136YrDc8orFywMHFX9Do7BT

### DIFF
--- a/docs/CONFIG_PARAMS.md
+++ b/docs/CONFIG_PARAMS.md
@@ -142,3 +142,59 @@ config:
 
 On restore, a manifest already declared in the config takes precedence over
 the checkpoint's (a warning is printed if they differ).
+
+## Observation Normalization (under `config:`)
+
+### `normalize_input`
+
+Enables online observation normalization: a `RunningMeanStd` tracks per-dimension
+mean/variance of the observations and the model normalizes every input with the
+current stats. Stats update on each training minibatch forward (train mode) and
+are frozen during rollouts and evaluation; they are part of the checkpoint.
+
+```yaml
+config:
+  normalize_input: True
+```
+
+**Default:** `False`. Applies to the PPO agents (`a2c_continuous`,
+`a2c_discrete`) and, when configured in `central_value_config`, to the central
+value network. SAC has its own `normalize_input` handling and is not affected by
+the warm-start parameter below.
+
+### `normalize_input_init_count`
+
+Seeds the sample count of the obs normalizer's zero-mean/unit-variance prior.
+With the legacy count of 1, the first training minibatch outweighs the prior
+thousands to one and the running stats jump straight to the batch stats; the
+policy recomputed under the shifted stats diverges from the rollout policy that
+collected the data (a large first-epoch KL — with `lr_schedule: adaptive` this
+can floor the LR at `min_lr` in epoch 1 before learning starts). A larger count
+turns that jump into a damped update.
+
+```yaml
+config:
+  normalize_input: True
+  normalize_input_init_count: 81920   # explicit prior weight (int or 8.2e4)
+  # normalize_input_init_count: 1     # legacy cold start
+  # omit or null                      # auto: one PPO epoch of samples (default)
+```
+
+**Default (`null`/absent):** `mini_epochs * horizon_length * num_actors *
+num_agents` — one PPO epoch of *counted* samples (the normalizer updates on
+every training minibatch, so its count accrues `mini_epochs`× the rollout size
+per epoch; the derivation matches that accounting). The prior then fades after
+roughly one epoch. Values below 1 are rejected (a 0 or negative count poisons
+the running stats).
+
+**Multi-GPU:** with pooled stats sync, the first merge sums every rank's seeded
+prior, so the effective prior weight is exactly one *global* epoch — consistent
+with the single-GPU behavior.
+
+**Scope and follow-ups:** this seeds the *input* normalizer of the PPO agents
+and the central value network. `value_mean_std` (`normalize_value`) has the
+same cold start and is a planned follow-up; SAC is unaffected. Note the warm
+start damps the first-epoch stat jump but does not change the update scheme
+itself — stats still drift within each epoch (updated per minibatch over the
+same rollout data); collection-time freezing à la VecNormalize would remove
+that drift entirely and remains a possible future change.
--- a/rl_games/algos_torch/a2c_continuous.py
+++ b/rl_games/algos_torch/a2c_continuous.py
@@ -33,6 +33,7 @@ class A2CAgent(a2c_common.ContinuousA2CBase):
             'value_size': self.env_info.get('value_size', 1),
             'normalize_value': self.normalize_value,
             'normalize_input': self.normalize_input,
+            'normalize_input_init_count': self.normalize_input_init_count,
         }
 
         self.model = self.network.build(build_config)
--- a/rl_games/algos_torch/a2c_continuous.py
+++ b/rl_games/algos_torch/a2c_continuous.py
@@ -64,7 +64,8 @@ class A2CAgent(a2c_common.ContinuousA2CBase):
                 'writter': self.writer,
                 'max_epochs': self.max_epochs,
                 'multi_gpu': self.multi_gpu,
-                'zero_rnn_on_done': self.zero_rnn_on_done
+                'zero_rnn_on_done': self.zero_rnn_on_done,
+                'normalize_input_init_count': self.normalize_input_init_count
             }
             self.central_value_net = central_value.CentralValueTrain(**cv_config).to(self.ppo_device)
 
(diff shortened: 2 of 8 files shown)


ViktorM (commented):
Reviewed the diff, the tests, and the interactions with what's on master (stats sync, checkpoints, players).

**Mechanism** — the fix is a scoped damping of the first-epoch jump, and that is where the damage happens (adaptive LR floored in epoch 1). Worth stating in the PR what it does *not* change: the obs normalizer still updates per minibatch in train mode, `mini_epochs` times over the same rollout data, so every epoch keeps a (decaying) train-vs-rollout drift and the count grows `mini_epochs`x faster than real samples — that's why "one epoch" has to be `mini_epochs * batch_size`. The canonical alternative (SB3 VecNormalize, CleanRL) updates the stats at collection time and trains with them frozen, which removes the drift entirely rather than damping its first step. Not asking for that here; `init_count` stays useful as a prior-strength knob either way. But please add a follow-up note.

**Evidence** — the KL table is at pinned LR; the claim that matters for users is about the *adaptive* schedule. Please attach LR + KL curves for the first ~20 epochs, cold vs warm, from one of the adaptive runs (myoHandPose LSTM is fine). Final rewards on the single-seed pairs are within noise except the transformer pathology, which is expected for a fix like this.

**Verified OK**
- Multi-GPU `pooled` sync: the first merge sums the seeded prior from every rank, so the prior weight becomes `world_size * mini_epochs * batch_size` = exactly one *global* epoch — consistent by accident. Add a one-line comment on the default derivation saying so; it is not obvious.
- Resume: `count` is in the state dict, restored checkpoints overwrite the seed; players build with count 1 and get it from the checkpoint. No action.
- `RunningMeanStdObs` per-key forwarding, int64 buffer, CI green.

**Needed before merge**
1. Docs: `docs/CONFIG_PARAMS.md` has no entry (it has none for `normalize_input` either — add both). This is a default-behaviour change for every PPO user; it also needs a row in the 2.0 migration notes (`normalize_input_init_count: 1` restores the legacy cold start) — I'll add that on our side.
2. Validate the value: `config.get(...) or 1` in models.py silently maps 0 to 1; raise on `< 1` in a2c_common instead, and cast floats from YAML (`8.2e4`) explicitly.
3. Tests: add one for the default derivation through the agent config (`None` -> `mini_epochs * horizon * num_actors * num_agents`) and one for `1` = legacy path. The RMS-level tests are fine.

**Suggested follow-up (separate PR)** — `value_mean_std` has the identical cold start and `normalize_value: True` is in nearly every config; same one-line seed. Central value likewise.

With 1–3 addressed: approve. Default ON is fine by me given the mechanism is strictly a damping of an epoch-1 artefact and the MLP results are neutral.

ViktorM:
Checked the head (2ecbc8c): none of the three review items is addressed yet, and the validation one is worse than it looked — verified live: `normalize_input_init_count: -16` silently poisons training (running stats go inf/-inf after one update), `0.9` gives count 0, and YAML `1e6`/`1.0e6` arrives as a *string* and crashes the ctor. Raise on anything < 1 and cast via `int(float(...))`.

Two additions to the earlier list:
- Central value doesn't get the warm start: `central_value.py` builds its `state_config` without the key, so the asymmetric critic's obs normalizer takes the exact first-batch jump this PR fixes. Thread it through — same one-liner as the actor path.
- SAC silently ignores the key (`sac_agent.py` net_config omits it) — either forward it or state PPO-only scope in the CONFIG_PARAMS entry.

Also worth a comment in code: the "one epoch" derivation is only correct because the `running_mean_std.eval()` calls before training are dead (set_train() re-enables updates per minibatch, so count accrues mini_epochs× per epoch). If that quirk is ever fixed, the derived prior becomes mini_epochs× too heavy — couple the two in a comment. And the multi-GPU pooled-sync interaction (per-rank priors sum to exactly one *global* epoch on the first merge) is fine but should be documented in the CONFIG_PARAMS entry.

Denys88:
Pushed 17d73cc addressing items 1–3 plus both additions from the follow-up comment:

1. **Docs** — `docs/CONFIG_PARAMS.md` now has entries for both `normalize_input` and `normalize_input_init_count`: the default derivation, the multi-GPU pooled-sync note (per-rank priors sum to one *global* epoch on the first merge), `1` as the legacy opt-out, SAC explicitly out of scope, and a follow-up note on `value_mean_std` and on collection-time normalization à la VecNormalize (this PR damps the first-epoch jump; it doesn't change the per-minibatch update scheme or its within-epoch drift).
2. **Validation** — extracted `resolve_obs_norm_init_count()`: `int(float(...))` so YAML scientific notation works whether it parses as float or string, and anything `< 1` raises an actionable error. Your live-verified cases are now tests: `-16` → raise, `0.9` → raise, `"1e6"` → 1,000,000. E2e-smoked: `8.2e4` in a real config trains; `0` raises at agent init. The `or 1` in models.py no longer swallows anything the agents pass (they only pass validated ints); it remains the permissive default for non-PPO builders.
3. **Tests** — default derivation (`None` → `mini_epochs * batch_size`), legacy `1`, scientific-notation strings, and the rejection paths.
4. **Central value** — `CentralValueTrain` now takes `normalize_input_init_count` and threads it into its `state_config`; both agents pass their resolved value, so the asymmetric critic's obs normalizer gets the same warm start.
5. **Derivation coupling** — the helper's docstring states the one-epoch derivation is tied to the per-minibatch update accounting (`set_train()` re-enabling updates), and must drop the `mini_epochs` factor if stat updates ever move to collection time.

**Adaptive-schedule evidence** (myoHandPose LSTM pair, `lr_schedule: adaptive`, first 20 epochs, same seed/code — cold is `init_count: 1`, warm is the auto default):

| epoch | cold KL | cold LR | warm KL | warm LR |
|---|---|---|---|---|
| 1 | 0.0772 | 1.1e-04 | 0.0375 | 2.3e-05 |
| 2 | 0.0112 | 7.6e-05 | 0.0101 | 1.1e-04 |
| 3 | 0.0113 | 7.6e-05 | 0.0082 | 1.1e-04 |
| 5 | 0.0112 | 7.6e-05 | 0.0104 | 7.6e-05 |
| 10 | 0.0109 | 1.1e-04 | 0.0121 | 1.1e-04 |
| 20 | 0.0104 | 1.7e-04 | 0.0110 | 1.1e-04 |

Honest reading: on the *LSTM* the adaptive LR recovers by epoch 2 in both runs — the flooring pathology is the transformer's (at `init_count: 1` its epoch-1 KL of 68.7 pins the LR at `min_lr` from epoch 1 for the entire run, with steady-state KL then sitting inside the adaptive dead band so it never climbs back). The LSTM's outcome gap on this env (+501 vs +228 final, deterministic same-seed pair) therefore comes from the early stat trajectory itself, not from LR flooring — the warm start halves the epoch-1 KL and the representation forms under stable normalization.

ViktorM (approved):

PR #372, opened 2026-08-30
MJLAB live viewer play with runtime command control + MicroDuck support

ViktorM:
Live viewer play for MJLAB tasks, and MicroDuck (pollen-robotics/microduck_rl) training support.

**Live viewer play**: `python -m rl_games.envs.mjlab_play --file <cfg.yaml> --checkpoint <ckpt.pth>`
- Native MuJoCo viewer, or viser in the browser when headless (`--viewer viser`; default `auto`).
- Velocity commands on the numeric keypad (KP8/KP2 forward/back, KP4/KP6 yaw, KP7/KP9 strafe, KP0 zero); every letter is already bound by the viewer's own key layers. Key presses are queued from the viewer thread and applied on the simulation thread.
- Restores any rl_games checkpoint through the standard player (obs normalization included); RNN state is zeroed on the viewer's full reset. Command control pins the velocity term's sampling distribution and re-asserts it every step, so mjlab's resample/standing-env machinery cannot revert it (0.0 deviation over 100 steps with mid-episode resets).
- `env_config: play: true` loads the task's registered play config; `player.env_config` replaces the training env block.

**MicroDuck**: `rl_games/configs/mjlab/ppo_microduck_velocity.yaml`, asymmetric actor-critic at their 4096x24 geometry, entropy 0, explicit LR band. Requires the mjlab-1.6 port of `mjlab_microduck` (upstream pins mjlab 1.3.0; port validated, publication pending). Results in docs/MJLAB.md: final return 128 +/- 6 over 4 runs vs 120.3 for rsl-rl (+6%); reaches their plateau at iteration ~290 (3.5 min) vs ~1330 (16 min) on their curve. Demo gif and comparison plot included.

**README**: the front page now leads with these results: a four-tile hero (MicroDuck speed lane, G1, Go1, WujiHand), the MicroDuck wall-clock comparison beside a clip of the shipped policy, the two cross-trainer MJLab plots side by side plus the Go1 rough-terrain plot, compact SAC and Isaac Gym sections, Play2Perfect added to the papers. Clips are rendered from the checkpoints with a following camera; MJLAB.md documents each clip's provenance and measured speeds.

**Also** (folded from #373/#374): `AdaptiveScheduler` default `max_lr` 1e-2 -> 1e-3 (configs that need more set it explicitly; go1/g1/lift_cube/ant pin their measured value); YAML scheduler bounds cast to float (`1e-5` loads as a string in PyYAML and crashed the first adjustment); Ant envpool config at 128 actors / minibatch 4096.

Validation: 257 passed / 1 skipped on master 7ab5039; CI green. Real stack: play-cfg construction, command control under resets, viser headless play and native-viewer play on a display with a trained Go1 checkpoint, 4096-env MicroDuck training (results above).


ViktorM:
Now also carries the Ant envpool config update (128 actors / mb 4096, two-seed measurement) and the AdaptiveScheduler default max_lr 1e-2 -> 1e-3 change (+ tests) — folded from the closed #373/#374.

Denys88 (commented):
Read the full diff and surrounding code (mjlab not installed here, so viewer-API claims that depend on mjlab internals are flagged as unverified). The viewer, the vecenv `play` flag and the tests are clean. What I want changed is the unrelated library-wide change riding along.

**Please split out: `AdaptiveScheduler` default `max_lr` 1e-2 → 1e-3** (`schedulers.py:20`, `a2c_common.py:359`, folded from #373/#374). In this tree 59 configs use `lr_schedule: adaptive` and only 12 set `max_lr` (7 on master + the 5 this PR adds). The other 47 stock configs — Isaac Gym, envpool, atari, mostly at `learning_rate: 3e-4` — previously let the KL-driven raise climb to 1e-2 and now saturate at 1e-3. Your own `ant_envpool.yaml:44-47` comment ("both seeds spent 121 / 135 epochs above 1e-3") shows a stock config that did. `test_perf_fixes.py:218-231` now locks the new default in. This may well be the right default, but it has nothing to do with a viewer feature and deserves its own PR/decision (and a release note, since `pyproject.toml` is already 2.0.0).

**Good bug fix**: casting `kl_threshold/min_lr/max_lr/lr_multiplier` to float (`a2c_common.py:355-359`). PyYAML reads `1e-5` as a string, and master's `AdaptiveScheduler.update` raised `TypeError` on the first out-of-band KL — master's own `CONFIG_PARAMS.md` example (`min_lr: 1e-5`) crashed.

**`mjlab_play.py`**

- `_pin_resample_distribution` (`:184-206`) does `setattr` on the live term cfg with only `getattr` guarded. If mjlab's command-term cfgs are frozen dataclasses this raises `FrozenInstanceError`. You validated on the real stack, so presumably not — just confirm.
- `restore_distribution` (`:166-182`) is never called by `run_play` (controller is only attached on the native path); dead code or wire it up.
- `--viewer auto` picks native only when `DISPLAY`/`WAYLAND_DISPLAY` is set, so on macOS it always falls back to viser even with a display. Matches the docs; a line in the `--help` text would save someone a puzzle.
- Thread handoff is fine (immutable tuple rebinding, `apply()` snapshots once). Import-time side effects fine (only `mjlab.viewer.native.keys` at module level, in try/except; test proves it with `sys.modules['mjlab'] = None`). RNN reset and obs-normalization restore paths verified against `players.py`.

**Configs**: g1/go1/lift_cube_yam only add `min_lr: 1e-6, max_lr: 1e-2` = no effective change. `ant_envpool.yaml` `num_actors` 64→128 and minibatch 2048→4096 = 2× frames per run with `max_epochs` unchanged — disclosed, fine.

**MicroDuck**: `ppo_microduck_velocity.yaml:66` (`Mjlab-Velocity-Flat-MicroDuck`) cannot resolve until the mjlab-1.6 port of `mjlab_microduck` is published, and the README now leads with MicroDuck results nobody can reproduce from this repo today. Either hold that part of the README until the port is up, or mark it clearly.

No hard-coded paths, no debug code. `player.env_config` "replaces wholesale" behavior is pre-existing (`player.py` unchanged), the docs just describe it now.

Merge order: overlaps #381 on `a2c_common.py`.

ViktorM:
Thanks for the read. Round 1 is on the branch (tip e1ab4cc, merged with master after #381; 330 tests pass).

**`max_lr` default 1e-2 → 1e-3: kept, with a reasoned no on splitting it out.** The value is not a side effect of the viewer feature, it is the fix for the failure this PR's config exists to avoid: on the free per-dim log-std head the KL-driven raise ran the rate up to the 1e-2 rail, the std grew exponentially and the policy collapsed (MicroDuck, documented in `docs/MJLAB.md`); the same ceiling sits an order of magnitude above every locomotion equilibrium we have measured (rsl-rl's own settles near 2.5e-4). A ceiling whose only effect is to permit that runaway is not a safe default, and 2.0.0 is the release to change it. It is disclosed in `CONFIG_PARAMS.md` ("Default `max_lr` changed 1e-2 → 1e-3 in 2.0.0") and will be in the release notes; the configs that measured a need for more set it explicitly (`ant_envpool.yaml` pins its measured value). If you would still rather see it as a separate PR for the record, say so and I split it, but the value stays.

**`mjlab_play.py`**
- `_pin_resample_distribution`: mjlab's `CommandTermCfg` is a plain (non-frozen) dataclass, verified on mjlab 1.6; the code now checks `__dataclass_params__.frozen` and raises a clear `TypeError` instead of a `FrozenInstanceError` if that ever changes.
- `restore_distribution` is now called from `run_play`'s `finally` before `env.close()`.
- `--viewer` help text explains the auto rule (macOS without `DISPLAY` gets viser; `--viewer native` forces the window).

**MicroDuck reproducibility**: the README now says, at the top of the MicroDuck results, that the runs need the mjlab-1.6 port of the task plugin and that it is not published yet, pointing to the port status in `docs/MJLAB.md`. Publishing the port is on our side and is the remaining blocker for anyone else reproducing the numbers.

**Recipe update**: the config now sets `normalize_value: true`; on the current code it lifts the final return from 128 / 130 to 136.1 / 138.3 / 139.0 on paired seeds (peaks 151 vs 143), the rest of the recipe unchanged. Numbers, mechanism and the negative results (step-KL scheduler, bigger critic, constant rate) are in `docs/MJLAB.md`. The description is rewritten to match the code (keypad keys, current numbers).

PR #373, opened 2026-08-31
Ant envpool config: 128 actors, minibatch 4096

ViktorM:
Config-only. Two-seed, same-window measurement on envpool 1.2.6: 128 actors / mb 4096 hits 6256-6411 reward at epoch 2000 vs 3253 for 64 actors at the same epoch (second seed; the first fell into the flipped-ant optimum and never recovered). Budget stays 2000 epochs (+4-7% only by 4000). HalfCheetah 2K probe showed +3.6% over 1K — its config left unchanged. Curves in the tuning TensorBoard (ant_4k_* runs).


ViktorM:
Folded into #372 per Viktor — single-commit PRs consolidated.

PR #374, opened 2026-08-31
AdaptiveScheduler: default max_lr 1e-2 -> 1e-3

ViktorM:
Default-change requested by Viktor: the legacy 1e-2 ceiling is ~10x above the controller's healthy equilibrium on GPU-sim locomotion (measured ~2.5e-4 on MicroDuck; the ceiling was only ever reached under the sigma-explosion pathology it then amplified — see the MicroDuck root-cause writeup). New default 1e-3 in both AdaptiveScheduler and the a2c_common fallback; docs updated; regression test added. Explicit per-task bounds remain recommended. Needs a release-notes row (default change for unset configs).


ViktorM:
Folded into #372 per Viktor — single-commit PRs consolidated.

PR #376, opened 2026-09-01
Release correctness: player vecenv fallback, CV normalizer prior, DDP unused-params + guard, drop OldGymWrapper

ViktorM:
Four confirmed defects from the release-preparation correctness review, plus one agreed removal. Full suite green (193 passed) including new regression tests.

- **Player**: registrations without an `env_creator` (envpool, pufferlib, plain GYMNASIUM entries — ~30 shipped configs) crashed `--play` with `KeyError: 'env_creator'`, a regression from the UV migration. `BasePlayer` now defaults `use_vecenv` to whether the registration lacks a creator; explicit `player.use_vecenv` still wins. *Downstream note*: a subclass overriding `create_env()` for an **unregistered** env name should set `use_vecenv: False` to keep the classic path.
- **Central value**: the obs-normalizer prior was forwarded from the actor unchanged, overweighting it by `actor_mini_epochs * num_agents / cv_mini_epochs` — exactly **27×** on the stock 27-agent SMAC config. `CentralValueTrain` now derives its default from its own geometry (`cv_mini_epochs * horizon * num_actors`; its state batch is per-env, not per-agent). Explicit values are read from `central_value_config.normalize_input_init_count` (the top-level key no longer leaks into the CV — documented in CONFIG_PARAMS).
- **DDP unused params**: `wrap_model_ddp` gains a `find_unused_parameters` passthrough (`central_value_config.ddp_find_unused_parameters`, documented), and the shipped central-value `TestNet` no longer builds a dead actor head — its discarded output made default DDP fail at iteration 2. *Note*: old checkpoints of the asymmetric test config carry `mean_linear.*` keys and won't strict-load; test-net only, nothing in-tree loads such checkpoints.
- **DDP guard**: `_TrackedDDP` marks `forward_seen` only after a completed, grad-enabled, sync-enabled forward — a no-grad/`no_sync()` forward installs no reduction hooks and previously armed the guard, letting a later raw-model backward step with unsynchronized gradients (two-rank test extended).
- **OldGymWrapper deleted** (maintainer decision): it could not wrap a genuine old-gym env (real `gym.spaces` fell through unconverted), advertised structured spaces while returning flattened values, and mislabeled old-gym timeouts as terminations. `convert_space` survives as a module function for the ManiSkill adapter.

Known follow-up (documented in-code, not fixed here): the CV one-mini-epoch stat freeze is a latent no-op (`train_critic`'s `self.train()` re-enables updates); the new prior deliberately matches actual accrual. Also observed once during review: `test_masked_rows_contribute_zero_gradient` flaked on a loaded box, unrelated to this diff (passes in isolation and on master).

Changes reviewed by a 3-agent adversarial pass; all findings addressed.


Denys88 (changes requested):
Reviewed against master ddc6a42, all findings adversarially verified. Verified clean first: the `_kl_for_lr_schedule` discrete refactor (global mode all-reduces identically, local mode safe via the rank-0 lr broadcast), the CV init-count derivation and its tests, and the `_TrackedDDP` no-grad guard. Four correctness/compat items and a style pass before approval:

**Correctness / compatibility**

1. **`test_network.py:26` breaks existing checkpoints.** Gating `mean_linear` on `not self.central_value` changes TestNet's `state_dict` keys; restoring any pre-PR asymmetric checkpoint (e.g. `test_asymmetric_discrete_mhv_mops.yaml`) hits strict `load_state_dict` at `a2c_common.py:955` and raises on the unexpected keys. Needs a compat shim or a load-path filter.

2. **`player.py:277` flips the default for unknown envs.** `_default_use_vecenv` returns True for any env lacking `env_creator` — including names entirely absent from `env_configurations` (via `.get(env_name, {})`). A downstream `BasePlayer` subclass overriding `create_env()` under such a name now crashes inside `vecenv.create_vec_env` before its override is consulted, and the traceback never suggests `use_vecenv: False`. Restrict the True default to envs actually registered with vecenv, or keep False when the name is unknown.

3. **`ddp_find_unused_parameters` is half-plumbed.** It reaches the central-value wrap only; the agent-side `wrap_model_ddp` call passes nothing, and a top-level key is accepted silently and ignored (keys aren't validated). Put the knob in `wrap_model_ddp` itself so both call sites get it — otherwise the only escape for an agent-side unused head is `flat_allreduce`, which this PR's own docs mark as scheduled for removal.

4. **`OldGymWrapper` removal breaks an unshipped promise.** Its docstring (still on master) says "kept one release for downstream users of the old contract", and no release ever shipped carrying it — tags end at v1.6.5, which shipped the class. v1.6.5 → 2.0.0 upgraders get ImportError with zero releases of grace. Keep a thin deprecation alias for one release, or declare the hard removal in the migration notes.

**Style — house-convention pass**

5. `central_value.py:304`: an 8-line comment documenting that the following `running_mean_std.eval()` is a known no-op. Delete the dead line instead of documenting it; keep a one-liner where the `mini_epochs` factor lives.
6. Several comments narrate PR history rather than behavior — the "exactly 27× on the stock 27-agent SMAC config" anecdote (already encoded in the test), "only the broken *default* forwarding was removed", "a regression from the UV migration". Commit-message content; keep the behavioral first line of each, drop the provenance.
7. The CV init-count resolution rule exists in three prose copies (two identical call-site comments in the byte-identical `cv_config` blocks + the helper docstring). Since `cv_config['config']` is already the merged `central_value_config`, the constructor can resolve the CV-scoped key itself — deletes both comments, shrinks the helper, and removes the function-local `a2c_common` import (no cycle exists).
8. `README.md:308`: the rewritten `use_vecenv` row drops the aligned-pipe column padding every neighboring row uses.

Minor, take or leave: `maniskill.py:55` runs `convert_space(remove_batch_dim(x))` where `remove_batch_dim` already rebuilds every space type; the new `ddp_find_unused_parameters` section in CONFIG_PARAMS.md is missing the `**Type:/Default:/Applies:**` header line its siblings use.

With 1–4 fixed and the comment pass applied: approve.

ViktorM:
1-3 and 7 in d875d32 / 904defa / 883e0d1 / dc29ffe: `TestNet` drops `mean_linear.*` on strict load in central-value mode (the parent`s strict load only reaches `_load_from_state_dict`); unregistered env names take the classic `create_env()` path, explicit `player.use_vecenv` wins, all 162 shipped configs swept; `ddp_find_unused_parameters` reaches both DDP wraps, the `central_value_config` key overrides; the CV init count is resolved in the constructor, helper and duplicate comments gone; the redundant `convert_space` call in maniskill dropped. Also on the branch: 0621ad0 (the red 3.12 job was `Runner.load` reseeding from `time.time()` in the test fixture, not a masking leak) and 940ba63 (unknown `schedule_type` raises; `standard_epoch` dropped). Every new test fails before its fix. 4, 5, 6 and 8 follow below.

ViktorM:
5, 6, 8 in 404e49e: the no-op `running_mean_std.eval()` in `train_net` deleted (normalizer count identical with and without it), the 27x anecdote dropped, the README `use_vecenv` row re-padded.

Denys88 (approved):
Verified on head 404e49e, item by item: (1) `_load_from_state_dict` drops `mean_linear.*` in CV mode — the right hook, it runs under the parent's strict load; (2) `_default_use_vecenv` is True only for registered creator-less envs, unregistered names keep the classic path so subclass `create_env()` overrides run; (3) `find_unused_parameters` lives in `wrap_model_ddp` and both call sites read it, CV key overriding top-level; (5) the no-op `running_mean_std.eval()` is deleted rather than documented; (6) history-narrating comments are gone; (7) the CV init-count rule has one home in the constructor and the module-top import is clean; (8) README row re-padded. Suite 201 passed / 1 skipped on this head, and the extras (masking seed pin, `schedule_type` validation) both check out.

On item 4: the deletion comment's finding that `OldGymWrapper` never actually converted real old-gym spaces settles alias-vs-removal in favor of hard removal — a one-release alias would only preserve broken behavior. Approving on that basis, with the expectation that the promised migration-notes line ("removed in 2.0; it was non-functional for genuine old-gym envs — use gymnasium's own compatibility wrappers") lands before the 2.0 tag.

ViktorM:
4 in 123bd58: `OldGymWrapper` stays importable as a stub whose constructor raises with the env-side gymnasium contract and the migration page. No alias: the wrapper never worked on a real old-gym env. Test added; suite 202 passed / 1 skipped.

Denys88:
Verified 123bd58 — the raising stub is the right close-out for item 4: importable name, actionable message, no broken alias shipped. Approval stands; all eight items closed. 🤖

PR #379, opened 2026-09-13
Masked variance: centered two-pass moments; legacy wrappers on the gymnasium 5-tuple contract

ViktorM:
Two items deferred from #377.

- **Masked moments** (`torch_ext.get_mean_var_with_masks`): the one-pass E[x^2] - E[x]^2 variance goes negative in float32 for nearly constant inputs; sqrt then turned every normalized advantage into NaN and the next update into NaN weights (a real agent on the validity-masked fixture reproduces it with constant advantages of 0.7 or 1.0). The path runs for every RNN rollout and every next-step-autoreset env, and the same helper feeds the masked `RunningMeanStd` and the masked diagnostics. Now a centered two-pass variance with the same clamps for 0/1 valid rows. Tests cover constant input, offsets up to 1000, multi-column pooled variance, degenerate masks, and `prepare_dataset` plus one optimizer step on a real agent.
- **Legacy wrappers on the gymnasium contract**: `wrappers.py` imports gymnasium but fifteen wrapper classes still returned obs-only resets and 4-tuples, while `RayWorker` unpacks `(obs, info)` and 5-tuples, so every route built on them was dead (`atari_gym`, `smac*`, the Cnn walkers, `TimeLimit`). All converted: `TimeLimit` expiry is `truncated` (and no longer clobbers an inner truncation's `time_outs`), `EpisodicLifeEnv` life loss is `terminated`, `MaxAndSkipEnv` stops on either flag, reset kwargs (seeds) forward through the stack. 38 tests on CartPole plus synthetic ALE-shaped and SMAC-shaped envs, including `RayWorker` round trips. The `dm_control` RAY route is retired (zero configs; the `gymnasium` and `envpool` routes cover them). The `CarRacing` route is keyed to the `-v3` id gymnasium actually registers.

Not verified here: real Atari (no ale_py) and SMAC (no StarCraft II); both are covered by synthetic stand-ins only. Open decision: pin SMAC v1 to rl-games 1.6.5 like Isaac Gym, or ship the routes API-correct but unverified.

Validation: 272 passed / 2 skipped on master 7ab5039.


Denys88 (commented):
Read the full diff and the surrounding code; checked old vs new masked-variance formulas numerically. Both halves are correct — the variance fix is a genuine NaN-regression fix and every consumer route (`RayWorker`, player) already speaks the 5-tuple contract. A few things before merge:

**Tests**

- `tests/test_legacy_wrappers.py:633-636` asserts `obs['state'].shape == (4, 7)` for `BatchedFrameStackWithStates(..., flatten=True)`, but the wrapper declares `state_space=(28,)` (`wrappers.py:342`) and `RayWorker.get_env_info` reports the unwrapped `(7,)`; `process_data` (`wrappers.py:370-372`) never stacks a 1-D state. The test now pins a shape that disagrees with both declared spaces, so SMAC + `frames>1` + central value is still unverified. Either fix the wrapper or don't cement the mismatch.
- `test_legacy_wrappers.py:682-685` comment says CarRacing "fell through to the unwrapped classic path" on master — it didn't; `carracing_ppo.yaml` hit a `KeyError` in `create_vec_env` (there is no fallback for creator-less names). With the `-v3` rekey it now actually builds the RAY/box2d chain (box2d is not a declared extra). Worth saying in the description.
- `EpisodicLifeEnv` new fallback (`wrappers.py:122-126`, no-op step ends the game → real reset) has no test; `FakeAtariEnv(n_lives=2, steps_per_life=1)` would hit it.
- nit: `tests/test_masked_variance.py:66-67` `sys.path.insert` is unnecessary (`tests/__init__.py` exists); `test_degenerate_masks_stay_finite` duplicates `test_ppo_masking.py:584`.

**Pre-existing, but this PR makes these wrappers live again**

- `MaxAndSkipEnv(use_max=False)` returns a view into the persistent buffer (`wrappers.py:192`); `FrameStack` then stores k references to the same memory, so all stacked frames show the latest observation. The new `use_max=False` tests never stack over it. A `.copy()` there would do.
- `FireResetEnv` (`wrappers.py:75-83`): if `step(2)` ends the episode the env is reset but the pre-reset obs is returned.
- Masked `RunningMeanStd` path feeds unbiased var while the unmasked path uses `unbiased=False`, and passes `input.size(0)` rather than the valid count as `batch_count`.

**Claim vs code**: "reset kwargs (seeds) forward through the stack" is true of the wrappers, but no in-tree caller passes them — `RayWorker.reset()` (`vecenv.py:132`) is bare and `RayWorker.seed` (`vecenv.py:125`) calls a non-existent `env.seed`. Only tests exercise it.

Verified fine: two-pass variance is pooled scalar and unbiased exactly like before (old form gives −3e-8 on constant 0.7 over 16 rows, new gives 0; random data with offset 100 matches `x[mask].var()` to float precision); all three callers expect a scalar; `TimeLimit` now `time_outs = truncated` and no longer clobbers an inner truncation; all 15 wrappers return 5-tuples / `(obs, info)`; `dm_control` RAY route removal is safe (depended on undeclared `dm2gym`, nothing references it).

Merge order: overlaps #381 on `torch_ext.py` and #368 on `env_configurations.py`.

PR #381, opened 2026-09-19
WujiHand: official recipe and results; exact KL, autocast-safe policy math, consistent value targets, max_sigma, kl_reference, diagnostics

ViktorM:
**WujiHand: official recipe and results.** `ppo_wujihand_reorient.yaml` uses a global exploration std (`fixed_sigma: true`, softplus, floor 0.2) with `entropy_coef 0`, wider actor/critic (1024/512/256, 1024/1024/512/256) and a **step-KL adaptive learning rate**: the scheduler reads KL(after step || before step) on the same minibatch, averaged over a mini-epoch (`kl_schedule_source: optimizer_step`, `schedule_type: standard`, target 0.002, band 5e-5..2e-4). Three seeds: 20.1 / 19.1 / 19.5 goal reaches per episode vs 16.9 for the vendored rsl-rl fork at the same 1.64B-frame budget (17.4 for the reference trainer at rl_games' widths), reaching the reference's final score in 1.30–1.44 h vs 2.1 h, equal action smoothness. Same recipe with a fixed 1e-4: 19.6 / 18.9 / 18.9; with the legacy reference-KL scheduler: 18.8 / 18.2 / 18.9. `docs/MJLAB.md` has the table, the plot, the mechanism behind the scheduler choice (with the recipe's rate response: fixed 5e-5 / 1e-4 / 1.5e-4 / 2e-4 = 18.5 / 19.9 / 19.2 / 16.8) and a short stability primer. On 2 GPUs (2× frames per iteration): **21.1**, reference score in 1.02 h (legacy scheduler on 2 GPUs: 20.6 / 1.27 h). The comparison plot and the gif are from the new recipe.

**Fixes (all users)**
- `policy_kl` is the exact diagonal-Gaussian KL in at least fp32.
- With `normalize_value`, old value predictions and returns are normalised with the same running statistics.
- Under `mixed_precision`, `mu` and the raw std head are fp32 before the distribution math.

**Behaviour changes for existing configs**
- Exact KL. The old `policy_kl` added 1e-5 inside the log and to the variance; for identical policies it reported, per action dimension, −0.00012 at σ = 0.2, −0.0005 at 0.1, −0.002 at 0.05, −0.012 at 0.02, −0.045 at 0.01. A 20-dim policy converged to σ = 0.05 read its KL 0.04 too low, five times the default `kl_threshold` of 0.008. With the exact value, adaptive-LR runs whose std has converged that far will see the rate reduced where it used to hold or rise. There is no compat switch; if such a run regresses, raise `kl_threshold` by the amount above.
- `normalize_value`: old value predictions in the unmasked branch are now normalised with the same running statistics as the returns, consistent with the central-value dataset. The `clip_value` window and explained variance shift slightly.
- `mixed_precision`: the distribution ops already ran in fp32 under autocast; the practical effect of the promotion is `bound_loss` / `reg_loss` on `mu` in fp32 and fp32 `mus` / `sigmas` in the result dict.

**New, opt-in**
- `kl_schedule_source: optimizer_step` (used by the Wuji config): the adaptive scheduler reads the KL of the optimizer step itself, from one extra no-grad forward per minibatch; `lr_multiplier`; applied-rate stats (`info/lr_mean|min|max`, `info/lr_at_min|max_fraction`, `info/scheduler_kl`). Default source unchanged.
- `max_sigma`: smooth ceiling on a state-dependent std, validated at build time, read by the `actor_critic` and `resnet_actor_critic` builders.
- `kl_reference: rollout`: adaptive-LR KL measured against the rollout policy.
- `use_diagnostics` also logs the std tail (min/mean/max), |mu| and |advantage| max, log-ratio max, the signed std score, the tail fraction, and matched-sample post-step KLs (`kl_step`, `kl_post_ref`, `kl_post_ref_max`; rank 0, not for recurrent policies).

Documented in `CONFIG_PARAMS.md`; 298 tests pass. The investigation itself is internal and not in the tree.


Denys88 (commented):
Read the full diff against master and the surrounding code; ran the 7 new test files against master too (38/48 fail there, so the suite discriminates). Core changes are correct — the fixes I'd like are disclosure and docs, plus two small code nits.

**Behavior changes that are not opt-in**

1. **Exact KL changes adaptive-LR dynamics for every existing continuous config** (`rl_games/algos_torch/torch_ext.py:78-92`). The formula is right, but the old epsilon bias was much larger than the "−0.0023 at σ=0.2" in the description once policies converge to small σ. Per action dim the old value was low by: σ=0.1 → 0.0005, σ=0.05 → 0.002, σ=0.02 → 0.012, σ=0.01 → 0.045. A 20-dim policy at σ=0.05 reported KL ≈0.04 too low — 5× the default `kl_threshold` of 0.008. So after this lands, existing Isaac-style adaptive configs will see the LR halve where it previously held or rose. Please state this explicitly in the description / release notes (or offer a compat switch).
2. `normalize_value` users get slightly different `old_values` (and hence `clip_value` window / explained variance) from the unmasked-branch fix at `a2c_common.py:1424-1431, 1726-1732`. Direction is correct and consistent with the central-value dataset — just worth a line.
3. `mus`/`sigmas` become fp32 under `mixed_precision`. No consumer depends on bf16, fine. Note that `softplus/exp/log/pow/sum` were already in the autocast fp32 list, so the "autocast-safe policy math" part is mostly cosmetic — the real effect is `bound_loss`/`reg_loss` on `mu` in fp32.

**Code nits**

- `models.py:250`: `span <= 0` is validated on every forward; move it to `network_builder.py:313` (config time). Only `A2CBuilder` reads `max_sigma`; `A2CResnetBuilder` silently ignores it (same as `min_sigma` today).
- No test reads `max_sigma` through the builder — all cap tests use `SimpleNamespace`, so the yaml key plumbing is untested. One builder-level test would close that.
- `test_continuous_agent_supplies_learner_policy_and_rollout_logprobs` executes the post-step forward but asserts nothing about `kl_step`/`kl_post_ref`.
- `policy_kl` has no guard for `p1_sigma == 0` (old code had +1e-5). Only reachable if `exp` σ underflows without `min_sigma`; probably fine, flagging it.

**Docs vs code**

- `docs/CONFIG_PARAMS.md:55-62` lists 6 diagnostics metrics; `diagnostics.py:51-60` logs 13 (`sigma_score_pos/neg/mean`, `tail_frac`, `kl_step`, `kl_post_ref`, `kl_post_ref_max` missing). The same paragraph says the metrics are "not post-update measurements", but `kl_step`/`kl_post_ref` are computed after `trancate_gradients_and_step()`.
- `docs/CONFIG_PARAMS.md:70`: "nearly unchanged well below the cap" is overstated — with floor 0.2 / cap 1.0: 0.3→0.29, 0.5→0.42, 0.8→0.54.
- `docs/MJLAB.md:154-155`: the key is `diagnostics/policy/sigma_max/{miniepoch}`.
- `README.md:14-15`: "1.6 h instead of 2.1 h" is the best of 1.56/1.86/1.60 h; the PR text itself says 1.6–1.9 h.

Verified fine: `kl_reference: rollout` (dataset `update_mu_sigma` skipped, per-row reference stable since `PPODataset` slices in fixed order, default preserved); diagnostics post-step forward (`norm_obs` returns a new tensor, stats frozen/restored, rank 0 only, RNN excluded); no new syncs in the hot loop with diagnostics off.

Merge order: this overlaps #379 on `torch_ext.py` and #372 on `a2c_common.py`.

ViktorM:
Thanks for the careful read. Everything below is in d58c377 unless marked otherwise.

**Behaviour changes that are not opt-in**

1. Exact KL: agreed, and your per-dimension numbers check out (old value for identical policies: `log(1 + 1e-5) + σ² / (2(σ² + 1e-5)) − ½` → −0.00012 at σ 0.2, −0.0005 at 0.1, −0.002 at 0.05, −0.012 at 0.02, −0.045 at 0.01). The PR description now has a "Behaviour changes for existing configs" section with that table and the consequence for adaptive-LR runs whose std has converged below ~0.05, plus the mitigation (raise `kl_threshold` by the bias). Proposal: no compat switch. The old number was a bias, not a definition, and a switch would keep two KL scales alive in the config space. Alternative if you would rather have one for a release: a `legacy_kl_epsilon: true` key (default off) that restores the old formula verbatim, removed in the next release; it is a five-line change and I can add it here.
2. `normalize_value` / `old_values`: noted in the same section.
3. fp32 under autocast: reworded. The promotion matters for `bound_loss` / `reg_loss` on `mu` and for the dtype of `mus` / `sigmas` in the result dict; the distribution ops were already on autocast's fp32 list.

**Code nits**

- `span <= 0` moved to build time: `read_sigma_bounds()` in `network_builder.py`, called from `A2CBuilder` and now also from `A2CResnetBuilder`, so the resnet path no longer silently ignores `min_sigma` / `max_sigma`. The per-forward check is gone.
- Builder-level test added: `max_sigma` read through `ModelBuilder` from the yaml keys and applied in the model forward (raw head 500 → std 0.998, below the cap); `max_sigma <= min_sigma` rejected at build.
- `test_continuous_agent_supplies_learner_policy_and_rollout_logprobs` now asserts `kl_step`, `kl_post_ref`, `kl_post_ref_max` are finite and positive and that max ≥ mean.
- `p1_sigma == 0`: proposal is to leave it as is. σ comes from `exp` / `softplus` of a finite head, the `exp` path has `logstd_bounds` and `min_sigma`, and a clamp inside `policy_kl` would reintroduce a bias in exactly the small-σ regime the fix targets. Alternative: `p1_sigma.clamp_min(1e-12)` inside `policy_kl`, a no-op for any std the builders can produce and a guard against a NaN if one ever underflows; say the word and I add it.

**Docs vs code**

- `CONFIG_PARAMS.md`: all 13 policy diagnostics listed, in two groups (in-batch pre-step; post-step forward on rank 0 through the raw module with the normaliser frozen, not for RNN). The "not post-update measurements" sentence is gone.
- `max_sigma`: "nearly unchanged" replaced by the numbers (floor 0.2 / cap 1.0: 0.3 → 0.29, 0.5 → 0.42, 0.8 → 0.54; exact only at the floor).
- `MJLAB.md`: `diagnostics/policy/sigma_max/<mini_epoch>`.
- README hours: since your review the shipped recipe moved to a fixed 1e-4 (40942be): 19.6 / 18.9 / 18.9 reaches, 1.4–1.5 h to the reference's score; README and MJLAB carry those numbers and the adaptive-band results stay in the table as a row. The PR description was still describing the band; updated to match.

**Merge order**: fine either way. If #379 / #372 go first I rebase this on top; if this goes first, the `torch_ext.py` / `a2c_common.py` overlap is small enough that I can do the rebase on their side too.

Denys88 (approved):
Re-reviewed da69c28. Everything from the first pass is addressed (KL table + consequences in the description, `read_sigma_bounds` at build time incl. resnet, builder-level `max_sigma` test, post-step diagnostics asserted, docs fixed). Ran the touched test files on this head: 119 passed.

Also read the `kl_schedule_source: optimizer_step` addition that came in after my review: default `reference` preserves the old signal exactly, RNN/discrete rejected at init, BatchNorm buffers restored around the matched forward, `.clone()` before the in-place all-reduce, and the extra forward only runs when the source is selected or diagnostics are on. Fine to ship as opt-in.

Decisions: no compat switch for the KL — agreed, it was a bias, the description now says so. `p1_sigma == 0`: leave it.

One nit, not blocking: `a2c_common.py:1717-1718` re-reads `min_lr`/`max_lr` from the config with its own defaults for `lr_at_min/max_fraction`; that duplicates the scheduler's defaults and will drift from #372's `max_lr` change. Reading `self.scheduler.min_lr`/`max_lr` when the scheduler is adaptive keeps the two in sync.

Approving. Rebase over #379 (`torch_ext.py`) whichever lands first.

ViktorM:
Scope update since the last round: the PR now also carries the learning-rate scheduler work that was on a side branch (`kl_schedule_source: optimizer_step`, `lr_multiplier`, applied-rate stats, tests, config docs), because the WujiHand config ships with it. The scheduler reads KL(after step || before step) on the same minibatch and acts once per mini-epoch on the mean; on three seeds it scores 20.1 / 19.1 / 19.5 vs 19.6 / 18.9 / 18.9 for the fixed 1e-4 rate and 18.8 / 18.2 / 18.9 for the legacy reference-KL scheduler, same smoothness; 21.1 on 2 GPUs (reference score in 1.02 h). Default behaviour for everyone else is unchanged (`kl_schedule_source: reference`). `docs/MJLAB.md` has the mechanism and the recipe's rate response; `CONFIG_PARAMS.md` has the calibration rule. Tip is 6d8b199; 298 tests pass.

